PR #471, opened 2026-07-25
fix(merge-gate): tier the denylist so its exception stops being the norm

BaseInfinity:
## The problem, measured

The flat `DENYLIST_PATTERNS` fired on **12 of the last 25 merged PRs (48%)**. A gate whose exception path is taken half the time trains bypass habits rather than caution.

Twice in one session it produced **worse outcomes than no gate**:
- **#465** — no sanctioned path for a mechanical rebase, so a clearance artifact was hand-written asserting a review that hadn't happened for that SHA
- **#468** — stalled a fully-verified merge behind a bypass the agent harness then denied

And `MERGE_CLEARANCE_SKIP=1` was **all-or-nothing**: escaping a benign doc hit also disabled the CI `validate` check, the net-test-deletion check, and every clearance-artifact check. There was no way to say *"yes this touches the doc, here's why that's fine — now still verify everything else."*

## The fix (Fable xhigh design, ROADMAP #478)

**`HARD_DENY`** — live enforcement + release machinery. **Cannot be acknowledged**; a human runs it, by design.
```
^.github/workflows/   ^hooks/   ^.claude/   ^scripts/merge-pr.sh$
```
`hooks/` is not merely shipped product — `.claude/settings.json` runs `codex-gate-check.sh` and friends from `$CLAUDE_PROJECT_DIR/hooks/`, so a PR gutting it weakens the pipeline that makes CERTIFIED mean anything. `.github/workflows/` defines `validate` **and** holds publish secrets, and a `pull_request` workflow runs the PR branch's own definition — so a PR editing `ci.yml` would pass its own gutted gate.

**`ACKABLE_DENY`** — prose that steers but doesn't mechanically enforce.
```
^CLAUDE_CODE_SDLC_WIZARD.md$   ^skills/sdlc/SKILL.md$   ^cowork/skills/sdlc/SKILL.md$
```
Proceeds only when the clearance artifact names **each matched path** with a reason and `"user_confirmed": true`. This is `ACK_DENYLIST_ONLY` — CI, test-deletion, CERTIFIED, `round>=2`, SHA freshness, and the review-file check **all still run**.

**`CHANGELOG.md` delisted** — pure dead weight. It appeared only on release PRs that also bump `package.json`'s version, which the separate version-field check already blocks. `^.github/workflows/release` removed as subsumed.

## Effect on the measured sample

| | |
|---|---|
| Hard (human runs it — correct) | **32%** |
| Ackable ("typo in a 250KB doc") | **16%** |
| Clean | **52%** |

## Tests

7 new tests, written **RED first**. Fable's named mutations verified:
- **M1** re-add `CHANGELOG` → fails ✅
- **M2** move `^hooks/` to ackable → fails ✅
- **M7** drop the `user_confirmed` requirement → fails ✅

**M7 initially passed** — the test only checked that the string `user_confirmed` appeared, so a mutation hardcoding `ACK_CONFIRMED="yes"` slipped through while leaving the word in a comment. It now asserts the value is **derived** from the artifact. Caught by mutation testing before shipping.

**38/38** merge-gate tests. **Full suite: 61 files, 0 failing.**

## Deliberately not done

- **Section-scoping the wizard doc** — disqualified. GitHub's `pulls/N/files` API omits the `patch` field for large diffs, and the 250KB doc is exactly the file most likely to have it omitted: a silent no-op, false-negative in the dangerous direction. The better path is splitting that doc (ROADMAP #476/#471), which makes path-matching precise again for free.
- **Renaming the break-glass to something benign** — that would be adversarial evasion of the harness's safety classifier. A skip-everything override *should* look alarming.


BaseInfinity:
Superseded by #474, which shipped the same tiering plus the approval/bypass split and the full test suite. Closing to avoid a stale DIRTY branch.

PR #487, opened 2026-08-04
release: v1.92.0 — remove the Cowork Stop hook (#484)

### LINKED ISSUE #484: Stop hook: 4 false blocks in one session, and it dumps its whole prompt into the terminal each time
## Two defects, one hook

`hooks/hooks.json` / `cowork/hooks/hooks.json` `Stop` prompt hook. Both observed live on 2026-08-02, Opus 5 `high`.

### 1. It is unusably verbose

Every block echoes **the entire ~40-line prompt** into the terminal before the reason. The maintainer's words: *"shit seems verbose."* At four blocks in one session that is ~160 lines of the hook's own instructions rendered to a human who needs one sentence: what was left unverified.

The prompt is written for the evaluator, not the user, but the *whole thing* surfaces on failure.

### 2. The evaluator ignores its own instructions — four times, all exempt cases

This is the serious half. **The prompt is correct. The evaluator does not follow it.**

| # | What the turn actually did | Why it was exempt | What it did |
|---|---|---|---|
| 1 | Read-only greps, no code changed | "judge THIS TURN ONLY" | Blocked, asserting code changed |
| 2 | Prior turn had already run 64/64 and said so | "a response that summarizes work done earlier is NOT evidence this turn changed code" | Blocked for not verifying a *different* turn's work |
| 3 | Delegated to a background subagent | "Work that is still IN FLIGHT… Pending is not unverified" | Blocked — **its own reason read "This is in-flight work, not unverified work"** |
| 4 | Two background suite runs in flight, response said so | same exemption | Blocked, naming both task IDs as evidence they were "running" |

Cases 3 and 4 are the damning ones: the evaluator **correctly classified the case as exempt and blocked anyway.**

### Why this matters more than an annoyance

v1.90.0's headline fix added the in-flight exemption to this prompt (GH #477). `tests/test-stop-hook-terminates.sh:161` passes with its own caveat written into the message: *"presence only; semantics require live E2E verification, not yet run."*

**That verification has now happened, unplanned, and the fix does not work.** The text is present. The evaluator ignores it. We shipped the observable half of #477 and left the actual failure mode untouched — exactly what the test's caveat warned might be true.

### What to do

1. **Cut the prompt hard.** It is ~40 lines of nuanced prose sent to Haiku. Anthropic's own guidance is that specific, concise instructions are followed more reliably. The exemption list has grown with every incident; that growth may itself be the cause.
2. **Suppress the prompt echo on block.** Surface the `reason`, not the instructions.
3. **Replace the presence-only test.** Execute the evaluator against fixtures and assert the DECISION — a read-only turn, an in-flight turn, a code-change-with-no-test turn. Grepping the prompt for words cannot distinguish a working gate from a broken one, which is precisely how this shipped.
4. **Consider whether this belongs as a prompt hook at all.** Rules 1 and 4 of `AGENTS.md`'s Code Review Rules apply: if an instruction must hold, encode it mechanically rather than asking a model to honor prose.

### Do NOT

Soften the prompt further. Three rounds of softening produced a prompt that is already correct and still ignored.

## Related

- ROADMAP #490 (testing contract — the presence-only assertion is recorded there)
- GH #477 (the original loop this was meant to fix)
- GH #456 (Cowork prompt hooks not reliably enforcing — same mechanism, different symptom)

BaseInfinity:
## What

Removes the Cowork prompt-type `Stop` hook. It fired **12 times in one session and was wrong 11 of them** — blocking turns that changed no files, turns whose verification was already stated in the response, and five separate times quoting its own in-flight exemption before overriding it. It fired **zero** times during the v1.91.0 release work, which shipped ten real defects.

Three prior prompt rewrites each added exemption language the evaluator then ignored, so a fourth rewrite was not the fix.

Cowork now has **no completion enforcement**. `PreToolUse` and `UserPromptSubmit` remain. `cowork/README.md`, the wizard doc, and E2E Test 5c are corrected — they promised a hook that no longer exists.

Also deletes ~190 lines of tests asserting the hook's prompt *text*. They passed continuously while it was wrong 11 times in 12.

## Guard changes

- Restored the suite's `[ "$FAIL" -eq 0 ] || exit 1` tail, lost during the edit above — the suite printed failures and exited 0.
- Replaced a **vacuous control assertion** that accepted "produced output" OR "was silent" (every possible outcome), so nothing ever executed the script behind the allowlisted path. Now asserts exit 0 on a normal stop, since exit 2 blocks the user's response.
- The guard no longer enumerates load paths. One rule: the only `Stop` hook in the repo is the three exact commands invoking `codex-review-stop-check.sh`. Scans `.json`, `.yaml`/`.yml`, `.md` frontmatter; skips only `.git`.

## Accepted risk

This guards **accident, not an adversary**. It cannot stop someone who deliberately re-adds a hook, because that person can also edit the guard. Six review rounds each defeated it under an insider-adversary threat model — that model is unsatisfiable for a repo-local test, and further hardening is out of scope per AGENTS.md rule 7 (`tests/` ships to nobody). Logged here so it is not re-litigated.

## Verification

- 64/64 suites pass
- `claude plugin validate cowork/` passes
- Live Cowork E2E independently confirmed the hook was already inert there

Closes #484


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Merge-safety review by GPT-5.6 Sol at `high`. Scoped to *is this safe to merge*, deliberately NOT to further hardening of the repo-local guard (accepted risk, documented in the test header, CHANGELOG, and handoff).

Verified: only Cowork `Stop` removed, `PreToolUse`/`UserPromptSubmit` byte-equivalent; all shipped version markers at 1.92.0; no shipped doc still promises completion enforcement; both plugin manifests validate; CLI tests 97/97; npm publish dry-run succeeds; CI validated this exact SHA across 82 steps. Negative controls confirm the tests exit nonzero when the old Stop hook is restored or the effort anchor is changed, and that the `grep -F` fix did not weaken the guard.

Noted, not blocking: `ROADMAP.md` still carries a v1.90.0 marker — pre-existing and repo-local, does not ship.

```json
{"reviewer": "gpt-5.6-sol", "confidence": 97, "sha": "9c95e8e3cb78e24492d6650d43ad09a6491ca40b"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Merge-safety review by Fable 5. Independent of the Codex review above; same scope.

Verified against the true PR range (origin/main...HEAD, commits 01748ef + 9c95e8e): Stop hook cleanly deleted with valid JSON after edit; CHANGELOG 1.92.0 entry is strictly append-only with all 105 historical entries intact — no blanket-sed damage, which has bitten two prior releases; version at 1.92.0 in exactly 9 shipped files; the only remaining Stop hooks are the three allowed `codex-review-stop-check.sh` command entries; npm surfaces consistent. Confirmed the `grep -F` fix matches identically on BSD and resurrects an assertion GNU grep had silently neutralized, and that the guard still fails on a mutated anchor.

Reported confidence is **93**, below this gate's 95 threshold. Posted for the record rather than to clear — it does not, and should not, count toward the two-reviewer requirement.

```json
{"reviewer": "fable-5", "confidence": 93, "sha": "9c95e8e3cb78e24492d6650d43ad09a6491ca40b"}
```

PR #495, opened 2026-08-07
--user-approved: let the human say the thing HARD_DENY demands

BaseInfinity:
## What

`./scripts/merge-pr.sh <PR#> --user-approved "<reason>"`

HARD_DENY printed *"a human decides this one"* and gave the human no way to say so. Three times in one session the only route left was clicking merge in the GitHub UI — which bypasses the hook harness and leaves **no record in the repo**. PR #479 was merged exactly that way.

**Waives process gates:** both denylist tiers, the version-bump confirmation, the clearance-artifact checks.
**Cannot waive facts:** a red CI `validate`, or a PR that net-removes tests. Both reviewers attacked that split and could not defeat it.

Posts a durable override record to the PR — reason plus every waived path — **before** merging, and refuses to merge if that post fails.

It does **not** prove a human typed it. Nothing available to a local script can. It makes an override explicit and durable instead of silent.

## Four defects found while building it

- Cleared the **stricter** tier then blocked on the weaker one.
- `--user-approved --cross-model-cleared` consumed the next flag **as the reason** — a full override justified by the string `--cross-model-cleared`. Whitespace reasons passed too. Found by probing, not by my tests.
- The success line always claimed `clearance CERTIFIED round>=2 fresh, denylist clear` — a **false audit record** when the override skipped both.
- The reason went to stderr and evaporated. PR #494 was merged with this flag and left **zero durable trace** — exactly what it was built to replace.

## Also: #486, three low-risk edits

Fable and Codex cross-examined each other and both concluded `SKILL.md` should **not** split per model. These are the agreed items.

- Concision guidance at CLAUDE.md level. Anthropic's actual position (a reminder in long prompts is fine) kept separate from this repo's stricter local ban — the first draft overstated the evidence.
- Cross-model CI-log audit scoped to release/workflow/control-plane PRs. One recorded catch doesn't justify a per-push layer.
- Preflight no longer frames *fewer findings* as success; the recheck's scope limiter is paired with "report every defect".

The #486 regression test was itself vacuous — it grepped for the wording the fix deletes. Now positively anchored and verified against the reviewer's own mutation.

## Status

**REVIEWED, not CERTIFIED.** Both reviewers returned NOT_CERTIFIED round 1; every finding is fixed, but neither has seen the fixes.

`SKILL.md` byte-neutral at 19,988/20,000, cowork byte-identical. Suite 65/65, merge gate 60/60.

Closes #479. Part of #486.

Issue #491, opened 2026-08-05
Audit CI/CD: green pipelines certified a phantom watchdog, 7 hooks that hang forever, and a hook whose test encodes its own defect


BaseInfinity:
Prompted by the 2026-08-04 `/doctor` run and the Opus 5 / Fable 5 harness audit (#486). Every finding below was verified against the repo, not inferred.

CI is green. It was green through all of the following.

## What CI is already doing right (verified, so the audit doesn't re-litigate it)

- All 64 top-level `tests/*.sh` suites are invoked by `.github/workflows/ci.yml`. **Zero orphans.**
- Every test path CI references exists. No phantom test steps.
- `ci.yml` enumerating each suite as its own step is now documented as the authority (CLAUDE.md, `fd57dcd`).

So this is not "CI doesn't run the tests." It's that **the tests, and CI, cannot see whole classes of defect that shipped anyway.**

## Class 1 — phantom references in shipped docs (VERIFIED, has caused harm)

`CLAUDE_CODE_SDLC_WIZARD.md:2535` tells every consumer:

> "The wrapper `scripts/codex-review.sh` already has a 30-min stall watchdog (`STALL_SECONDS=1800`) as the real timeout control."

**Neither exists.** `scripts/codex-review.sh` is not in the repo, and `STALL_SECONDS` appears nowhere in `scripts/` or `hooks/`. A grep across shipped docs finds exactly one phantom `scripts/*.sh` reference — this one.

Not cosmetic: this session found **14 stale `codex` processes, oldest 3.5 days**, precisely because the watchdog everyone was told to rely on does not exist. Both cross-model reviewers independently repeated the claim as fact because it is in our own shipped documentation. One of them called it "the exact WebFetch-class error in a different costume."

**Proposed check:** every `scripts/*.sh`, `hooks/*.sh`, and `tests/*.sh` path named in shipped docs must exist on disk. Trivially mechanical, would have caught this the day it was written.

## Class 2 — hook robustness is untested, and the hooks ship

**Seven shipped hooks perform an unbounded read on stdin:**

- `hooks/sdlc-prompt-check.sh:45`
- `hooks/codex-gate-check.sh:16`
- `hooks/tdd-pretool-check.sh:24`
- `hooks/codex-review-stop-check.sh:23`
- `hooks/precompact-seam-check.sh:30`
- `hooks/token-spike-check.sh:24`
- `.claude/hooks/merge-gate-check.sh:31`

[...]

Issue #476, opened 2026-07-27
Recommend native install for Claude Code CLI (avoid sudo-npm footgun)


BaseInfinity:
## Context

The wizard never documents how to install the Claude Code CLI itself — `CLAUDE_CODE_SDLC_WIZARD.md` line ~100 references `npm i -g @anthropic-ai/claude-code@<new_version>` but that's internal maintainer tooling for the version-test benchmark procedure, not end-user guidance. There's no consumer-facing recommendation anywhere in the wizard docs.

Official docs (https://code.claude.com/docs/en/setup) explicitly label native install as **"Recommended"**:
- Native installs auto-update in the background; no manual `claude update` needed.
- npm installs can silently fail to auto-update if the global npm dir isn't writable.
- Docs explicitly warn: **do NOT use `sudo npm install -g`** — it causes exactly the permission/ownership problems that break future updates and uninstalls.

## What happened (live repro, 2026-07-27)

On a real machine, `@anthropic-ai/claude-code` had at some point been installed via `sudo npm install`, silently leaving `/opt/homebrew/lib/node_modules/@anthropic-ai/claude-code` **root-owned**. This caused:
1. `claude update` to fail with "Insufficient permissions to install update."
2. A subsequent `npm uninstall -g @anthropic-ai/claude-code` to fail with `EACCES` (can't rename a root-owned directory as a normal user).
3. Two `claude` binaries left on `PATH` simultaneously after switching to `claude install` (native) — the native one (`~/.local/bin/claude`) and the stale root-owned npm shim (`/opt/homebrew/bin/claude`), silently resolved by PATH order with no warning. If PATH order ever changed, the user would silently fall back to the old npm-pinned version.

None of this is a wizard bug — it's a real footgun a wizard-driven session hit while doing routine maintenance, and the wizard is exactly the kind of doc that should steer users away from it up front.

## Proposal

Add a short note somewhere sensible (README "Getting Started" and/or `CLAUDE_CODE_SDLC_WIZARD.md`'s prerequisites section) recommending:
- Install Claude Code via the native installer (`curl -fsSL https://claude.ai/install.sh | bash` / the Windows equivalents), matching the official docs' own "Recommended" guidance.
- Explicitly warn against `sudo npm install -g @anthropic-ai/claude-code`.
- Optionally, a one-liner for `claude doctor` as the way to spot conflicting installs (the official docs point there too: "Check for conflicting installations").

[...]

PR #497, opened 2026-08-07
fix(hooks,docs): three shipped defects, and five rounds of fixing the tests meant to catch them

BaseInfinity:
## Three defects that shipped to consumers

**1. `hooks/model-effort-check.sh` drained stdin with a bare `cat > /dev/null`.**
v1.94.0 exists to remove exactly this — a read-to-EOF that was observed alive at
10h19m against a 10-second timeout. Its SessionStart twin `token-spike-check.sh`
was converted correctly; this one was missed.

*Why it escaped:* `tests/test-hook-stdin-bounded.sh` already tested the behaviour,
but its hook roster was hand-listed and this hook was never in it. A hand-maintained
roster stops covering whatever nobody remembered to add, and reports success either
way. The roster is now derived from `hooks/hooks.json` at any depth, fails if it
drifts from the manifest, and fails rather than skipping when an entry doesn't resolve.

**2. `hooks/instructions-loaded-check.sh` shipped the footgun GH #476 is about.**
On every session start with an update available it told consumers to run a global
npm install. #476 reads as "we omit native-install guidance"; in fact we shipped the
opposite advice, from a hook that cannot detect which channel installed the binary.
Now channel-neutral, covering native/npm and package-managed installs.

**3. The docs promised a 30-minute stall watchdog that has never existed.**
`scripts/codex-review-with-progress.sh` loops on `kill -0` and enforces no timeout.
The first fix here was to repoint the sentence at the wrapper that *does* exist —
which `ROADMAP.md` had explicitly warned against, because renaming satisfies a path
check while leaving the promise just as false and harder to spot. Reverted.

## The distribution boundary

`npm pack --dry-run` ships 25 files. **None are from `scripts/`.** Five shipped
references pointed consumers at tools they never receive, including
`skills/sdlc/SKILL.md` mandating `scripts/merge-pr.sh`. The guard I first wrote
blessed all five, because it asked whether paths exist *on the maintainer's disk*.
It now asks whether the consumer gets them.

## Five review rounds

Neither reviewer certified until round 5. Findings: **8 → 6 → 3 → 2 → 0**.
Everything fixed was *demonstrated* by a reviewer, not merely suspected.

Two detectors were **deleted rather than widened** — a source-text `cat` regex that
missed `$(cat 2>/dev/null)`, `/bin/cat`, `mapfile`, `IFS= read`, and a one-item
variable denylist escapable by renaming the phantom. Both are the unwinnable
denylist shape ROADMAP #495(a) describes.

Round 4 surfaced a genuine **reviewer conflict**: one wanted bare `review` in the
watchdog scope; the other's must-not-fire case contains that word. Each was right
about its own case and blind to the other's. Resolved by scoping on the claim
(`reviews are bounded`) rather than the topic word, so both hold.

## Verification

- Suite **65/65**
- Mutation matrix spanning all five rounds: **5 must-fire cases fire, 4 must-not-fire stay silent**
- `skills/sdlc/SKILL.md` 19,993 bytes, byte-identical to its cowork twin
- Both reviewers **CERTIFIED** at `b591570`

## Known residual (non-blocking, both reviewers agree)

Deliberate-evasion paraphrases can still slip a regex pin; neither reviewer thinks
that space is closable this way. Follow-up filed: the mutation matrix currently
lives in prose and should be executable fixtures — rounds 2–4 each shipped a
detector regression that only reviewer probing caught.

Closes part of #491. Fixes the shipped half of #476.


BaseInfinity:

**Reason:** Maintainer standing instruction, given in-session: 'if codex and fable agree you dont need me'. Both reviewers CERTIFIED at b591570 after five rounds (findings 8 -> 6 -> 3 -> 2 -> 0), each verifying the other's must-not-fire case after a genuine reviewer conflict was resolved. CI validate green in 3m26s, and the new #476 test is confirmed PASSING on the claude-less runner it would previously have reddened. HARD tier applies because the diff touches hooks/.

- `hooks/instructions-loaded-check.sh (merge-evidence path)\nhooks/model-effort-check.sh (merge-evidence path)\nCLAUDE_CODE_SDLC_WIZARD.md (guidance path)\ncowork/skills/sdlc/SKILL.md (guidance path)\nskills/sdlc/SKILL.md (guidance path)\n clearance-artifact checks (.reviews/merge-clearance-497.json)\n`

Head: `b59157005fbfd440dc91ba68e13ae9c3654eace6`

PR #500, opened 2026-08-08
ci(rename): make the self-PR skip tolerant of both repo slugs

BaseInfinity:
Groundwork for renaming this repo to `claude-sdlc-harness`. Landed **before** the
rename so the rename itself cannot break anything.

## The defect this prevents

`pr-review.yml` skips the paid Claude review on this repo's own PRs:

```yaml
github.repository != 'BaseInfinity/claude-sdlc-wizard'
```

Both reviewers independently flagged the same failure: **a GitHub rename does not
rewrite that string.** The redirect covers web and git operations, not a literal
comparison inside a workflow. The moment the repo is renamed, the condition flips
true and the review this skip exists to prevent starts running on every self-PR —
burning quota with nothing to indicate anything changed.

Listing both slugs makes the rename a no-op for this gate **in either direction**,
so it can be flipped without a flag day and rolled back without a second one.

## Verification

- TDD RED observed first: the test demanded the new slug and failed for that reason
- Mutation: inverting `!=` to `==` still fails the negative control
- Suite 65/65

## What this does NOT cover — the step a redirect cannot do

**npm Trusted Publishing binds to the exact repository name.** `release.yml`
publishes via OIDC, and the publisher configured on npmjs.com names
`BaseInfinity/claude-sdlc-wizard` + `release.yml`. After the rename the OIDC claim
no longer matches, and **the next tagged release fails at publish** until the
publisher config is updated in the npm console. `package.json`'s `repository.url`
must be updated to match as well, since provenance verification checks it.

Codex live-tested the redirect claim rather than assuming it: marketplace `add` and
`update` both succeed through this repo's *previous* slug
(`BaseInfinity/agentic-ai-sdlc-wizard`) on Claude Code 2.1.221. So plugin installs
survive — but never create a new repo at either historical name, since that destroys
the redirect for every pinned install.

Part of the wizard → harness rename. Prose rename follows separately; it is not a
`sed` (`skills/sdlc/SKILL.md` lands at exactly 20,001 bytes against a 20,000 ceiling
if all 8 occurrences are replaced).

### DIFF
--- a/.github/workflows/pr-review.yml
+++ b/.github/workflows/pr-review.yml
@@ -31,8 +31,15 @@ jobs:
     # PRs (see SDLC skill), so the Claude PR review is redundant on self-repo.
     # Consumers using pr-review.yml in their own projects WILL run this normally —
     # the skip only fires when github.repository matches the wizard repo exactly.
+    # Both slugs are listed on purpose. A GitHub rename does not rewrite this
+    # string: the redirect covers web and git, not a literal comparison inside a
+    # workflow. With only the old slug here, renaming the repo would silently
+    # flip this condition true and start running the paid self-review on our own
+    # PRs — burning quota with nothing to indicate anything had changed. Listing
+    # both makes the rename a no-op for this gate, in either direction.
     if: |
       github.repository != 'BaseInfinity/claude-sdlc-wizard' &&
+      github.repository != 'BaseInfinity/claude-sdlc-harness' &&
       ((github.event.action == 'opened' && github.event.pull_request.draft != true) ||
        github.event.action == 'synchronize' ||
        github.event.action == 'ready_for_review' ||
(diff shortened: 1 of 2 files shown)


BaseInfinity:

**Reason:** Maintainer directed the wizard->harness rename this session and instructed me to execute it ('wtf you rename it', 'you can do all that right?'). This PR is the safety net that must land BEFORE the rename: both Fable and Codex independently identified pr-review.yml's literal repo-slug comparison as silently flipping true on rename and starting the paid self-review on our own PRs. TDD RED observed, mutation-verified, suite 65/65, CI validate green in 3m59s. HARD tier applies because the diff touches .github/workflows/.

- `.github/workflows/pr-review.yml (merge-evidence path)\n clearance-artifact checks (.reviews/merge-clearance-500.json)\n`

Head: `f983d49102f97e09773815c62001efef46694ea5`

PR #503, opened 2026-08-08
chore(rename): wizard → harness for the repo slug and brand, identifiers held

BaseInfinity:
The GitHub repo is now `BaseInfinity/claude-sdlc-harness`. This updates what points
at it and what calls it by name, and deliberately changes **nothing a consumer's
install depends on**.

## Renamed

- Repo slug in **23 files** — marketplace sources, plugin manifests,
  `package.json` `repository.url`, shipped skills, workflows, docs
- Brand phrase `"SDLC Wizard"` → `"SDLC Harness"` across **38 files**
- `package.json` description, now leading with *self-evolving* — the word the README
  has used all along, and which the GitHub description now matches

## Deliberately NOT renamed

Scope decided by Fable and Codex, consulted blind; both independently recommended
repo + prose only.

| Kept | Why |
|---|---|
| npm package `agentic-sdlc-wizard` | 641 weekly downloads; npm cannot rename in place and a new package starts at zero |
| Plugin IDs `sdlc-wizard`, `sdlc-wizard-cowork`, marketplace name | installs key on these — renaming forces everyone to reinstall and changes their skill namespace |
| CLI bin `sdlc-wizard`, `/claude-*-wizard` commands, `CLAUDE_CODE_SDLC_WIZARD.md`, `SDLC_WIZARD_CACHE_DIR` | functional identifiers, not brand |

There is precedent: this repo was already renamed once
(`agentic-ai-sdlc-wizard` → `claude-sdlc-wizard`) while npm stayed put, with no
fallout. Consumers touch exactly one identifier per surface.

Kept on the **old** slug on purpose: `pr-review.yml`'s dual-slug list and the test
that pins both (so the rename stays a no-op for that gate in either direction), plus
CHANGELOG and the archived Codex findings — those were true when written.

## One test needed a real fix, not a cosmetic one

`test_wizard_url_live` fetched `main` and asserted the title. Pinning the *new* title
made this change **unmergeable by its own gate**: the check reads published state from
inside the unpublished diff, `main` still serves the old title, `ci.yml` runs the
suite, and `merge-pr.sh` treats CI-green as **not waivable**. It now accepts either
brand for the transition window. The property worth asserting is that the URL serves
the wizard document — not that the title is one exact string.

## Out-of-band, already done

npm Trusted Publishing was re-pointed at `BaseInfinity/claude-sdlc-harness`
(`release.yml`, `npm publish` only). GitHub's redirect does not cover npm's OIDC
identity, so without this the next tagged release would have failed at publish.
Verified on npmjs.com. The `npm trust github` CLI returns a bodyless 400 on a payload
that passes its own validation — the web form is the supported path.

## TDD deliberately skipped

On the maintainer's explicit instruction: a one-time mechanical rename that will not
recur, verified after the fact rather than driven by a regression test nobody will
ever need again. Recorded here rather than skipped silently.

## Verified

- Suite **65/65**
- npm package, `bin`, and plugin IDs unchanged after the sweep
- `skills/sdlc/SKILL.md` still **19,993 bytes**, byte-identical to its cowork twin
  (it had zero brand-phrase hits, so the 20,000 ceiling never came into play)
- `repository.url` points at the new slug, which npm validates against the trusted
  publisher at publish time


BaseInfinity:

**Reason:** Maintainer directed the wizard->harness rename this session and approved shipping v1.95.0 ('yup'). Fable, used as decider per the maintainer's Fable-is-the-brain instruction, ranked shipping v1.95 as the single highest-leverage next action: the work is finished and unbanked, and every branch cut after the rename rebases painfully until it lands. CI validate green. Rename scope (repo+prose, identifiers held) was decided by Fable and Codex consulted blind, and npm Trusted Publishing has already been re-pointed at the new slug and verified. HARD tier applies because the diff touches .github/workflows/ and skills/.

- `.github/workflows/cc-version-drift.yml (merge-evidence path)\n.github/workflows/pr-review.yml (merge-evidence path)\n.github/workflows/release.yml (merge-evidence path)\n.github/workflows/weekly-api-update.yml (merge-evidence path)\nhooks/hooks.json (merge-evidence path)\nhooks/instructions-loaded-check.sh (merge-evidence path)\nCLAUDE_CODE_SDLC_WIZARD.md (guidance path)\ncowork/hooks/hooks.json (guidance path)\n clearance-artifact checks (.reviews/merge-clearance-503.json)\n`

Head: `dec550ef1a338299c3015568ef73049fbc7ada65`

Issue #504, opened 2026-08-08
Which model should drive: Opus 5 calling Fable, or Fable driving and delegating to Opus subagents?


BaseInfinity:
Maintainer, 2026-08-07:

> "fable is the brain here...opus the driver, self review with fable, then finally if
> that passes self review with codex — and we should investigate whether its better to
> run as Opus 5 and then call Fable with advisor/subagents OR have Fable be driver and
> call out to Opus subagents"

## What is now settled, and shipped

The **order** is decided and codified in `CLAUDE_CODE_SDLC_WIZARD.md`:

1. **Fable decides** the approach — it is the primary thinker, not a rung you escalate
   to when stuck
2. **Opus implements** it
3. **Fable reviews** the implementation
4. **Codex** runs last as the adversarial cross-model gate

Fable appears twice deliberately: once as the brain before code exists, once as the
reviewer of what got built. Codex is last and singular — a gate, not a second opinion
to average with the first.

## What is NOT settled — this issue

**Who sits in the driver's seat.** Today: Opus 5 drives and calls out to Fable per
decision. The inverse is untested: Fable drives and delegates implementation to Opus
subagents.

The argument for inverting is not cosmetic. In the current arrangement the stronger
reasoner is consulted **on request** — which means the driver has to *notice* it needs
help. Every time it fails to notice, the decision is made by the weaker reasoner and
nobody finds out. Inverting puts continuous reasoning in the seat that makes decisions
continuously, and makes delegation the explicit act instead.

Live evidence that the current shape has this failure mode, from the session that
produced this issue: the driver picked a rename scope, created four milestones, and
formed a recommendation **before** consulting Fable at all. When Fable was finally
asked, it overruled the milestone cut, caught a dependency inversion between #491 and
#498, and moved the maintainer's stated goal (Opus 5 alignment) from third to next.
None of that required information the driver lacked. It required judgement the driver
did not pause to borrow. That is exactly the failure the "consult on request" topology
produces.

## What to measure — not vibes

Do not settle this by argument; both arrangements sound reasonable, which is why it is
still open.

- **Cost per unit of work.** Fable-as-driver runs the expensive model continuously.
  Opus-as-driver runs it per consultation. Quantify before assuming the inverse is
  affordable.
- **Decision quality.** Use the existing E2E scoring harness (`tests/e2e/`) rather
  than inventing a new one: run the same scenarios under both topologies and compare
[...]

Issue #489, opened 2026-08-04
skills/sdlc/SKILL.md is at its 20,000-byte ceiling — every edit now starts by degrading something else


BaseInfinity:
## The defect

`skills/sdlc/SKILL.md` is **19,993 bytes** against a **20,000-byte** ceiling enforced by `tests/test-audit-session-load.sh`. Seven bytes of headroom.

That is no longer a budget. It is a mechanism that converts every future edit into a forced trade, and the trades have started producing real defects.

## Evidence from one session, 2026-08-03

Three separate content defects, all caused by shaving bytes to fit:

1. **Two invalid JSON examples.** Compressing the dialogue-loop example produced `{"finding","action":...}` — not parseable. A doc example that cannot be copied is worse than no example. *Found by Codex, not by me.*
2. **A value dropped from an instruction.** `"commit_sha": "<git rev-parse HEAD>"` became `` `commit_sha` `` — the key with no value, so the instruction no longer said what to write. *Found by Codex.*
3. **A contract-guarded table reworded.** While fitting the #483 change I edited a row of the canonical Confidence Check table, which must stay byte-exact across `skills/sdlc/` and `cowork/skills/sdlc/`. *Caught by `test-doc-consistency.sh`* — the only one of the three a test caught.

Also lost and restored in the same pass: the `/goal` contract's **time** bound (leaving only the turn bound).

The #483 change alone required **four** rounds of trimming. Three of those touched sections unrelated to the change being made.

## Why this is structural, not housekeeping

- The file is the **operational contract** every consumer repo inherits. Content lost here is guidance nobody receives.
- The pressure is invisible at review time: a diff adding two lines and removing four elsewhere looks like tidying, not deletion.
- It selects against exactly the wrong content — the trims land on prose that has no test guarding it, because guarded text fails loudly and unguarded text does not. **The least-protected guidance is the most likely to be cut.**
- `cowork/skills/sdlc/SKILL.md` must stay byte-identical (`test-cowork-drift.sh`), so there is no room to diverge.

## `/doctor` does NOT solve this — checked

Two independent reviews of CC 2.1.220 confirmed `/doctor` finds **unused** skills/MCP/plugins versus their context cost and can disable them. It does **not** trim skill bodies. The "rightsize your skills" claim circulating in an article is overstated; no changelog entry through 2.1.220 supports it.

## Options, unranked — this needs a decision, not a default

[...]

Issue #485, opened 2026-08-03
Replace 2,443 lines of hand-rolled doc guards with declarative consistency checking


BaseInfinity:
## The problem, measured

`tests/test-doc-consistency.sh` is **2,443 lines and 170 test functions**. A large fraction encode one pattern: *"this value must be stated consistently across the shipped files."*

Written out by hand, per value, per file. Examples currently in the tree:

- `test_ai_setup_lanes_reviewer_is_gpt56`
- `test_readme_reviewer_is_gpt56`
- `test_wizard_doc_reviewer_is_gpt56`
- `test_skill_files_reviewer_is_gpt56`
- `test_agents_md_reviewer_is_gpt56`
- `test_claude_md_reviewer_is_gpt56`

**Six functions, one fact.** The same shape repeats for Sonnet 5's default effort, Opus 1M framing, model-pin recommendations, and now driver and reviewer effort.

## Why this is worth fixing, with tonight's evidence

Changing one value — the Opus 5 driver effort default from `xhigh` to `high`/`medium` — took **three cross-model review rounds and produced nine separate contradictions** across shipped files. Every round followed the same script: fix the files I remembered, add or update an anchor guard, declare it done, reviewer finds more.

The reviewer's diagnosis was exact: *"The guard passes because it does not cover this file."*

That is not carelessness that better attention fixes. **A guard enumerating locations cannot detect a location nobody enumerated.** The ninth instance was found only after the guard was widened to scan all shipped files for the assertion *shape* rather than checking a list of anchors — and it found one that two independent reviewers had missed.

Same class of defect as ROADMAP #495 (the reviewer-effort denylist arms race) and #490 (tests asserting text rather than behaviour). Three symptoms, one cause: **the check is written per-incident instead of per-property.**

## Proposed shape

A declarative manifest of shipped facts plus one generic checker. Roughly:

```yaml
- fact: reviewer_model
  must_say: ["GPT-5.6", "Sol"]
  must_not_say: ["GPT-5.5", "GPT-5.4"]
  scope: shipped            # resolved from `npm pack --dry-run`, not `files`
  except: [historical-citation-markers]

- fact: driver_effort_default
  must_say: ["`high`", "`medium`"]
  must_not_assert: "Opus 5 .{0,40}`xhigh`"
  allow_qualified: ["escalat", "planner", "not the default"]
```

[...]

PR #505, opened 2026-08-08
fix(cowork): the prompt gate must tell a bypass from a justified exception, and Fable decides

BaseInfinity:
Two things, both about a rule that was right in general and had no path for the case
where it wasn't.

## 1. The Cowork prompt gate blocked a reasoned engineering call

Live incident, 2026-08-07. The maintainer wrote:

> "do the prose rename now, dont TDD something like this, it will most likely only ever
> happen once so just verify after"

The gate denied it and **ended the turn outright, with no retry.** That is not a bypass:
it states a reason (a one-time mechanical rename that will not recur) and an alternative
(verify afterwards). The safeguard was being *substituted*, not removed.

The gate could not tell the difference because it only asked "does this ask to skip
planning, testing, or review". Any sentence containing a skip was a bypass, regardless
of what was offered in its place.

Confirmed ours before touching anything: the blocking prompt was read out of the
installed plugin cache (`~/.claude/plugins/cache/.../sdlc-wizard-cowork/1.93.0`) and is
byte-identical to `cowork/hooks/hooks.json` here.

The prompt now draws the distinction, still denies bare skips, and adds a tie-break it
lacked: **when ambiguous, allow.** This gate ends the turn with no retry, so a wrong
denial costs the user their whole turn while a wrong allow costs nothing — the TDD and
review gates downstream still fire. Asymmetric consequences deserve an asymmetric
default.

**This is the third gate this session that had no sanctioned exception path.** The merge
gate got `--user-approved` for the same reason. A rule people learn to route around is
worse than one with a recorded exception.

## 2. Fable is the primary thinker, not a rung you escalate to

The shipped guidance framed Fable as somewhere to go when stuck, and as a cadence slot
("Fable during design"). Both read as advisory. The contract is stronger, and the loop is
now stated in order:

**Fable decides → Opus implements → Fable reviews → Codex gates.**

Fable appears twice on purpose: the brain before code exists, the reviewer of what got
built. Codex is last and singular — an adversarial gate, not a second opinion to average
with the first.

Recorded as explicitly unsettled: this runs Opus as driver calling out to Fable. The
inverse (Fable driving, delegating to Opus subagents) is untested and may be better —
tracked as #504. The doc now says "the one that has been used, not the one that has been
measured" rather than implying it was chosen on evidence.

This rule went to per-user memory first, which breaks the skill's own Memory Audit
Protocol: *a process rule saved only to memory is a /sdlc gap.* Worth naming why —
`skills/sdlc/SKILL.md` is at **19,993 of 20,000 bytes**, so the ceiling is actively
pushing process rules out of shipped docs into private memory. That is #489 doing damage
beyond "edits are annoying".

## 3. ROADMAP.md asserted two things that were false

Under **"Start here next"** — the first thing a cold session reads — it claimed row #485
was implemented and awaiting review on `fix/485-single-tier-merge-gate` with `HARD_DENY`
deleted. Verified: that branch is **0 commits ahead of main** and carries nothing, and
`HARD_DENY` is alive in `merge-pr.sh` with **5 occurrences**. It had been sending cold
sessions to work that does not exist, on a branch that is empty.

It also said "Open PRs — none" while #503 was open. The integrity test enforces the
*form* of bullets in that block, so a claim of "none" has no bullets to inspect. The fix
is not a better checker — whether a PR is open is the tracker's fact and no offline test
can verify it. ROADMAP now links the PR list instead of restating it.

## Verification

- Suite **65/65**, rebased onto the post-rename `main`
- Every new assertion RED first. Two of my own were vacuous and caught while writing
  them: `reason` matched the JSON field name already in the prompt, and
  `alternative|verify` matched a word elsewhere in the text. Both tightened, then each
  clause confirmed load-bearing by deleting it and watching the test fail.
- The Fable assertion initially failed against *correct* prose because the doc says
  `Fable **decides**` and the pattern wanted a literal space — same markdown-weld class
  as the #491 splitter bug, caught here by the test rather than a reviewer.

Part of v1.95.0. Refs #504.

### DIFF
--- a/CLAUDE_CODE_SDLC_WIZARD.md
+++ b/CLAUDE_CODE_SDLC_WIZARD.md
@@ -2482,7 +2482,11 @@ Before presenting approach, STATE your confidence:
 
 Low confidence does **not** mean "ask the user." It means "escalate," and the user is the third rung, not the first. Ask a human only for what no model can settle.
 
-1. **Fable** — `advisor()` before plans; if the advisor is unavailable, spawn a Fable subagent at `high`.
+1. **Fable — the primary thinker, not a rung you escalate to.** Fable **decides** design, priority and sequencing; you implement its call. You are the coder. Reach for it *before* an approach exists, via `advisor()`, or a Fable subagent at `high` while the advisor is server-side disabled. **The tell that you got this backwards:** you have already chosen an approach and are asking Fable to grade it. Prompt it to decide — "where you would normally say consider X, say do X" — rather than to critique. Codex is a different job entirely: the **adversarial check** that tries to break a decision already made, run before commit or merge. Fable thinks, Codex attacks; do not collapse them into "get a second opinion".
+
+   **The loop, in order.** Fable decides the approach → Opus implements it → **Fable reviews the implementation** → only once that is clean, Codex runs as the final cross-model check. Fable appears twice on purpose: once as the brain before code exists, once as the reviewer of what got built. Codex is last and singular — it is the adversarial gate, not a second opinion to average with the first. Sending work to Codex before Fable has reviewed it wastes the expensive check on defects the cheaper one would have caught.
+
+   **Open question, not yet settled:** this topology runs Opus as the driver calling out to Fable. The inverse — Fable as driver, delegating implementation to Opus subagents — has not been tested and may be better, since it puts the stronger reasoner in the seat that makes decisions continuously rather than on request. Do not assume the current arrangement is optimal; it is the one that has been used, not the one that has been measured.
 2. **Codex `high`** — when Fable can't close the gap, or when a second, adversarially-framed opinion is what's needed.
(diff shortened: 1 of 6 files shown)


BaseInfinity:

**Reason:** Maintainer approved shipping v1.95.0 ('yup'). This PR carries the Cowork prompt-gate fix that unblocks the maintainer (the gate ended his turn for a reasoned one-time TDD waiver), the Fable-decides contract he dictated this session, and two false claims removed from ROADMAP.md. CI validate green in 3m36s, suite 65/65 rebased onto post-rename main. Every new assertion was RED first and each clause confirmed load-bearing. HARD tier applies because the diff touches cowork/hooks/ and shipped guidance.

- `CLAUDE_CODE_SDLC_WIZARD.md (guidance path)\ncowork/hooks/hooks.json (guidance path)\n clearance-artifact checks (.reviews/merge-clearance-505.json)\n`

Head: `96fb0e64fcf41ff2cd2193d7a4c34a6a37874d1e`

Issue #493, opened 2026-08-05
Version-marker guard: three markers sat at v1.90.0 for three releases, and three attempts at a guard were each proven vacuous


BaseInfinity:
Split out of the v1.94.0 release after three failed attempts to make it non-vacuous. The release ships without it; the three stale markers it was written for are fixed by hand.

## The defect class

The v1.94.0 release found **three live version markers frozen at v1.90.0**. They had survived the v1.92.0 and v1.93.0 releases untouched:

- `ROADMAP.md` — `**Last release: v1.90.0**`
- `ROADMAP.md` — `Dogfooded, v1.90.0` (the living-tracker row `SDLC.md:123` names as a release marker)
- `ROADMAP.md` — `` `main` is at v1.90.0 in `cowork/.claude-plugin/plugin.json` ``

Each was found by a separate round of cross-model release review. **None was caught by any of the 65 test suites.**

**Why a release sweep structurally cannot find them:** bumping N → N+1 works by searching for N. A marker stuck at N-3 does not contain N, so it is invisible to the sweep that is supposed to maintain it. The only reliable check is the inverse — assert each marker **equals** `package.json`.

Related, same root: `test_claude_md_reviewer_is_gpt56` pinned a hardcoded line number and false-passed against a decoy line (fixed in `fd57dcd`); five sibling call sites still do.

## Why the first three attempts failed — read this before writing the fourth

Each attempt passed a green suite and was proven vacuous by an independent reviewer.

**Attempt 1 — nine named markers, missing extraction treated as success.**
Replacing any single extractor regex with `NEVERMATCH` still produced a green 111-test suite. **Eight of nine markers were silently unchecked.** The guard reproduced, inside itself, the exact defect it existed to catch.

**Attempt 2 — report missing extraction as a failure; add marketplace coverage; mutate a copy.**
Better, and all seven named labels verified RED individually. Still broken in two ways:

- The marketplace loops filter values through a SemVer regex *before* counting, so `NO-VERSIONS-FOUND` fires only when **all** values vanish. Changing each of the five marketplace versions to `"NEVERMATCH"` individually left the suite green **every time**.
[...]

PR #506, opened 2026-08-08
release: v1.95.0 — claude-sdlc-harness, and three defects that were shipping

BaseInfinity:
## What ships

The repo is now **`BaseInfinity/claude-sdlc-harness`**. The npm package stays
`agentic-sdlc-wizard`; plugin IDs, CLI bin and slash commands are unchanged. Those are
what an install depends on — renaming them would force every consumer to reinstall for
no functional gain. The old repo URL still redirects, and **npm Trusted Publishing has
been re-pointed at the new slug**, which GitHub's redirect does *not* cover (verified on
npmjs.com; without it this release would have failed at publish).

Carries #495, #497, #500, #503, #505.

### Three defects that were shipping to every consumer

| Defect | Why it escaped |
|---|---|
| A hook could block forever on stdin — the exact unbounded read v1.94.0 existed to remove | its test roster was hand-listed and this hook was never in it |
| We shipped the install footgun our docs warn about | the #476 test greps `hooks/` only |
| Docs promised a stall watchdog that has never existed | nothing checked that a named mechanism was real |

Plus the distribution boundary: `npm pack` ships no `scripts/`, yet five shipped
references pointed consumers at tools they never receive.

### Gates that had no exception path

`--user-approved` on the merge gate, and the Cowork prompt gate now distinguishing a
bypass from a justified exception. Three gates this session blocked the maintainer with
no sanctioned way through; that pattern is the finding, not each instance.

### Fable decides

Shipped guidance framed Fable as a rung to escalate to. Now stated in order: **Fable
decides → Opus implements → Fable reviews → Codex gates.** Whether Fable should drive
instead is recorded as untested (#504), not settled.

## Verification

- Suite **65/65**
- **Eleven version markers** had to move together. The suite caught three I missed — the
  cowork marketplace entry, the update skill's example, and the SDLC.md table. That is
  #493's defect class caught by tests rather than by review, which has not happened
  before.
- Historical `v1.94.0` references in ROADMAP and two test comments left in place
  deliberately — they describe what was true then.
- Every constituent PR was cross-model reviewed before merge; #497 ran **five rounds**
  until both reviewers certified.

## After merge

Tag `v1.95.0` and push it — that triggers the npm publish under the new trusted
publisher.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -6,7 +6,7 @@
   },
   "metadata": {
     "description": "SDLC enforcement plugins for AI-assisted development",
-    "version": "1.94.0"
+    "version": "1.95.0"
   },
   "plugins": [
     {
@@ -15,13 +15,18 @@
         "source": "github",
         "repo": "BaseInfinity/claude-sdlc-harness"
       },
-      "description": "SDLC enforcement for AI agents — TDD, planning, self-review, CI shepherd",
-      "version": "1.94.0",
+      "description": "SDLC enforcement for AI agents \u2014 TDD, planning, self-review, CI shepherd",
+      "version": "1.95.0",
       "author": {
         "name": "Stefan Ayala"
       },
       "category": "productivity",
-      "tags": ["sdlc", "tdd", "code-quality", "testing"]
+      "tags": [
+        "sdlc",
+        "tdd",
+        "code-quality",
+        "testing"
+      ]
     },
     {
       "name": "sdlc-wizard-cowork",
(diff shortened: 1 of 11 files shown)


BaseInfinity:

**Reason:** Maintainer approved shipping v1.95.0 ('yup'). Release PR: version bump across 11 markers, CHANGELOG, and the rename. CI validate green in 3m59s, publish-dry-run green, suite 65/65. Every constituent PR (#495 #497 #500 #503 #505) was cross-model reviewed before merge; #497 ran five rounds until both Fable and Codex certified. npm Trusted Publishing already re-pointed at the new slug and verified on npmjs.com. HARD tier applies because the diff touches shipped skills and plugin manifests, and it is a version bump.

- `CLAUDE_CODE_SDLC_WIZARD.md (guidance path)\n package.json version bump (release-adjacent confirmation)\n clearance-artifact checks (.reviews/merge-clearance-506.json)\n`

Head: `0ff3ece7606954b4dc5dfe1f84c653915bf99b22`

PR #507, opened 2026-08-08
docs(readme): the E2E scoring pipeline runs here, not in the consumer's repo

BaseInfinity:
See commit message. Rolled into v1.96.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,15 +221,24 @@ GENERATED FILES (in your repo)
   - .claude/skills/*/SKILL.md
   - .claude/settings.json
   - CLAUDE.md, SDLC.md, TESTING.md, ARCHITECTURE.md
+
+        (that's everything you get — the arrow below leaves your repo)
+
         |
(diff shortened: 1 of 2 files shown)


BaseInfinity:

**Reason:** Maintainer asked for this to roll into the next milestone rather than ship alone ('just roll it into next milestone/release'), and to keep moving. README's architecture diagram implied the E2E scoring pipeline runs in the consumer's repo; verified against package.json files and npm pack that tests/e2e/ ships to nobody. Same distribution-boundary defect class as #497. TDD RED first, test added so it cannot drift back, CI validate green in 4m5s, suite 65/65. Assigned to v1.96.0 so it ships with that release. HARD tier applies because the diff touches shipped docs and tests.

- ` clearance-artifact checks (.reviews/merge-clearance-507.json)\n`

Head: `cd7892efda4aabedfec92e42ac1bd758d6484bd6`

Issue #486, opened 2026-08-03
Audit shipped guidance against the Opus 5 prompting guide — and decide whether it needs splitting per model


BaseInfinity:
Promoted from ROADMAP #493 so it lives where work is tracked (see #482).

## Why

Anthropic publishes model-specific prompting guides for [Opus 5](https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/prompting-claude-opus-5), [Fable 5](https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/prompting-claude-fable-5), and [general best practices](https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/claude-prompting-best-practices). We ship guidance written before them. **The Fable guide has never been read at all**, despite Fable being this repo's primary reviewer and advisor.

Four places the Opus 5 guide contradicts what we ship (verified against the files, not paraphrased):

1. *"If your prompt contains explicit verification instructions… remove them: instructions like these cause over-verification on Claude Opus 5."* — `skills/sdlc/SKILL.md` makes `self_review` a **critical** rubric item.
2. *"do not use subagents to verify or double-check your own work."* — that is our entire cross-model review protocol.
3. *"If your review prompt says 'only report high-severity issues'… it may report less."* — our recheck prompt says "Do NOT expand the surface."
4. *"Avoid instructing re-checks it already performs."*

## Practitioner report, 2026-08-02 (r/ClaudeCode, "Reworking CLAUDE.md for Opus 5")

A user hit the symptom this repo has been chasing — Opus 5 slow, and making implementation mistakes caught later by code review. Rather than reverting to 4.x, he fed the Opus 5 guide plus his instruction file back to Opus 5 and asked what to change. His reported result:

> "The file's biggest cost driver is that it instructs Opus 5 to do things it already does — and **Opus 5 compounds those instructions rather than ignoring them.**"

Three things this adds:

- **A mechanism, not just a symptom.** *Compounding* is the testable claim. It predicts redundant instructions cost more on Opus 5 than on 4.x — which would make our verification-heavy prose actively expensive rather than merely inert.
- **A subagent-delegation cap** was among the recommendations. We spawn Fable and Codex constantly, and our shipped guidance encourages it.
- **The finding with the largest blast radius: guidance has DIVERGED between Opus 5 and other models.** His audit flagged instructions redundant for Opus 5 that Copilot still needed, and concluded shared instruction files now need splitting per model.

[...]

PR #509, opened 2026-08-08
fix(sdlc): delete the same-model self-review instructions, keep the measurement

BaseInfinity:
Closes the last remaining edit in #486.

## The evidence

Anthropic's Opus 5 guidance says explicit verification instructions cause
over-verification. This repo's own record is blunter: `/code-review` reported **64/64
green three times in one session** while an independent model found real P1s each time —
including a shipped hook proven silently dead and a guard proven to be reading nothing.

**Same-model self-review has zero recorded unique catches here.**

## Delete the instructions, keep the measurement

**Deleted** — TodoWrite checklist item (skill + wizard doc), the `## Self-Review Loop`
section, and `hooks/sdlc-prompt-check.sh`'s per-prompt `/code-review` directive. That last
one is the worst surface for instructed verification: it fires every turn and compounds.

**Kept** — the `self_review` rubric row, still 1 point, still *"read back the files/diffs
you modified"*. It loses only its **critical** flag; `tdd_red` becomes the sole must-pass
and `check_critical_criteria` drops its self_review clauses.

### Why the row stays — this was the reviewer's correction to the issue as filed

#486 said to delete the row. Deleting it saves ~80 bytes and **invalidates every
longitudinal baseline**: `baselines.json` is all `max_score: 10`, and the golden scores,
CUSUM history and score analytics are calibrated on that total.

A read-back is cheap, evaluator-verifiable hygiene. What was wrong was *instructing* it
every turn and *failing a run* for missing it — not measuring it.

## Cross-model review is untouched

The over-verification finding targets **same-model** self-checking and does not transfer.
Cross-model review is the only layer with a record of catching real defects here.
`/code-review` keeps the one job it has evidence for: preflight input that reduces what
the cross-model reviewer has to find.

The wizard doc keeps the detail, including that a self-review pass is still reasonable for
**Sonnet-class drivers** — model-conditional guidance belongs in the on-demand doc, not in
a skill shipped to every model.

## Byte ceiling

| | |
|---|---|
| `skills/sdlc/SKILL.md` | **19,993 → 19,445** |
| Headroom | **7 bytes → 555** |
| cowork twin | byte-identical |

Edits stop being forced trades — which is half of #489 solved as a side effect.

## Coupling the plan missed, caught by the suite

`test_enforcement_coverage_score` counts documented sections having TodoWrite items, and
self-review was 1 of 12. The denominator drops to 11 rather than the item coming back:
enforcing a checklist step for prose that no longer exists would be enforcing nothing.

## Verification

Both new assertions RED first — `test_sdlc_hook_self_review_reference` inverted to assert
the directive is *gone*, and the ordering test replaced with an absence assertion (an
ordering check on a deleted section can only pass vacuously). Suite **65/65**.


BaseInfinity:

**Reason:** DUAL CROSS-MODEL CERTIFICATION on head 06a0f02 — merging under the maintainer's standing policy that both reviewers agreeing is the merge authorization and he is the deadlock-breaker only. Fable: CERTIFIED. Codex: CERTIFIED. Both verdicts bound to this exact SHA, both reviewers re-ran their own adversarial mutations rather than accepting summaries. Seven rounds, findings 8->6->3->2->3->2->1->0. CI validate green on head; 65/65 under both the normal locale and LC_ALL=C.UTF-8; 8/8 e2e. Recorded honestly: this is --user-approved because no --dual-certified path exists yet (GH #511 is the design to fix exactly this).

- `hooks/sdlc-prompt-check.sh (merge-evidence path)\nCLAUDE_CODE_SDLC_WIZARD.md (guidance path)\ncowork/skills/sdlc/SKILL.md (guidance path)\nskills/sdlc/SKILL.md (guidance path)\n clearance-artifact checks (.reviews/merge-clearance-509.json)\n`

Head: `06a0f0285dd31fe0da209b44f46b60f0a209a26b`

PR #517, opened 2026-08-08
feat(merge-gate): dual cross-model certification is merge authorization (#511)

### LINKED ISSUE #511: Dual cross-model certification should BE the merge authorization — the gate cannot currently express it
Maintainer, 2026-08-08, stating it as policy rather than preference:

> "if fable and codex agree with the merge then you don't need me, you only need me if
> there's back and forth and nothing gets resolved where they both aren't happy with the
> work... that's true automation and only bringing me in when needed and proper agentic
> /sdlc on all layers. I trust fable and opus 5 together and our iteration in /sdlc that
> you only need HITL when the system breaks or they don't agree multiple times with 95%
> confidence and the system has to come to a halt"

## The gap

`scripts/merge-pr.sh` has two clearance paths:

| Flag | What it does |
|---|---|
| `--user-approved "<reason>"` | records a human decision, waives the denylist tiers |
| `--cross-model-cleared` | satisfies **only** the denylist finding, not the other checks |

**There is no "both reviewers certified this SHA" clearance.** So the standing policy above
has to be routed through a per-PR human decision — which is precisely the thing it exists to
eliminate. Every merge on 2026-08-07/08 (#500, #503, #505, #506, #507) recorded "the
maintainer said yes" when what actually happened was the maintainer had *already* said yes,
once, as policy.

The mechanism is forcing a lie of framing into the audit trail. That is worse than an
inconvenience: the override records are supposed to be the durable evidence of *why*
something merged, and right now they misattribute a standing policy to a per-PR judgement.

## Why HARD tier existed, and why dual cross-model genuinely changes it

HARD tier covers the merge-evidence chain — `.github/workflows/`, `hooks/`, `.claude/`,
`scripts/merge-pr.sh`. The reasoning was sound: a PR that edits the machinery deciding
whether it may merge cannot clear itself. Self-produced clearance is circular.

**Dual cross-model certification breaks that circle in a way self-review never could.** Two
models that did not write the code, run blind to each other, each able to refuse. That is a
categorically different evidence class from "the author says it is fine" — and this repo has
the record to prove it: PR #497 took **five rounds** before both certified, and every round
found real defects the author had missed.

## The hard part — do not skip this

**The clearance artifact is written by the agent.** Nothing stops a sufficiently motivated
driver from writing `{"status":"CERTIFIED"}` without running anything. The current
`handoff.json` + `commit_sha` staleness check (#437) raises the cost but does not close it.

So the design question is not "should dual agreement suffice" — the maintainer has decided
that. It is **what makes a forged dual certification hard enough to be worth trusting**, given
a local script cannot verify a model actually ran. Candidate directions, to be judged not
assumed:

- require the raw reviewer output files, non-empty, with per-round distinct content
- require `round >= 2` for HARD paths (a real dialogue, not a first-pass rubber stamp)
- require the two verdicts to come from *different* reviewer identities recorded in the artifact
- server-side required check, which is the only actually-tamper-resistant answer and is
  currently out of scope (already logged as a Research Parking Lot item)

## What must stay non-waivable regardless

- **Red CI `validate`** — already non-waivable, stays that way
- **Net-removed test files** — same
- These are mechanically verifiable and no amount of model agreement should clear them.

## Deliverable

1. A `--dual-certified` path in `merge-pr.sh` that accepts verified dual cross-model
   clearance for HARD-tier paths, with an honest success line distinguishing what was
   *verified* from what was *attested*.
2. The shipped guidance updated so consumers inherit the policy, not just this repo:
   dual cross-model agreement is merge authorization; the human is the deadlock-breaker.
3. `--user-approved` stays, for the genuine deadlock case and for anything the reviewers
   cannot settle.

## Decide before building

Route the design through Fable (decider) and Codex (adversarial check) — including whether
the anti-forgery bar above is achievable locally at all, or whether this should wait for a
server-side check. **Do not build a clearance path that merely looks rigorous.** That would
be the worst outcome here: the appearance of a gate with none of the substance, on exactly
the paths that define the gate.

Refs #479 (approval/bypass split), #437 (commit_sha staleness), #497 (five-round proof that
cross-model review catches real defects).

BaseInfinity:
Closes #511.

## What this changes

`scripts/merge-pr.sh` gains `--dual-certified`: two distinct reviewers' visible, SHA-bound
`YES` clearances at >=95 confidence now clear a **HARD-tier** path — the merge-evidence chain
(`.github/workflows/`, `hooks/`, `.claude/`, the merge script itself), which previously only
`--user-approved` could.

The maintainer stated it as standing policy: *"if fable and codex agree with the merge then you
don't need me."* The gate could not express that, so every HARD-tier merge recorded a **per-PR
human decision** for what was really one policy decision made once. That is a framing lie in
the exact records that exist to explain why something merged, which is worse than the friction
it caused.

## Why dual agreement is admissible where self-review is not

Two models that did not write the code, run blind to each other, each able to refuse. That is a
different evidence class from "the author says it is fine" — and this PR is its own evidence:
**three rounds, 11 findings, none disputed**, including one P0 that both reviewers found
independently.

## What it verifies, and what it does not

**Verified mechanically:**

- CI `validate` green across *every* run of that name — paginated, both page envelopes, with a
  queued or in-progress run counted as not-green
- no net-removed test files
- no `package.json` version bump
- clearance artifact CERTIFIED at round >= 2 bound to the head SHA — **not waived** here,
  unlike `--user-approved`
- the **executing** merge script and the working-tree redirect hook byte-match `origin/main`

**Stated, not glossed:** both clearances come from the same `gh` token, so this is **attested,
not authenticated**. A new `.github/workflows/` file can still mint a green required check;
nothing local closes that, and the compensating layer is that both reviewers read that file in
the diff. The self-integrity check cannot defend against its own deletion. All of this is in
the header, the success line, the posted record, and ROADMAP.md.

**Never clears:** red CI, deleted tests, version bumps. `--user-approved` remains, now scoped
to what it always meant — reviewer deadlock, and releases.

## Two logically separate changes, called out rather than buried

1. The `--dual-certified` path (above).
2. **A pre-existing CI defect**, found while building it: the `validate` lookup took `head -1`,
   so a green check-run shadowed a red one of the same name. It also could not see the
   `"conclusion": null` a queued run carries. Both fixed; the extraction moved from regex to
   `jq`, which also retired a regex branch that never matched a live payload.

## Review record

| Round | Codex | Fable |
|---|---|---|
| 1 | 5 findings | 5 findings (3 overlapping) |
| 2 | 2 new P1s | CERTIFIED + 2 P3 |
| 3 | **CERTIFIED** | *(see thread)* |

The findings worth naming, because they say something about the method:

- **The self-integrity check bound the wrong file.** It compared cwd-relative paths and never
  looked at `$0`, so the file verified and the file executing could differ — and the failure
  message *steered you into the gap* ("run from a clean main checkout" invites re-running the
  absolute path you already have). Both reviewers found it independently.
- **The fixture was demonstrating the hole it existed to guard**, committing two-line `echo
  gate` dummies as the "pristine" gate while executing the real script by absolute path.
- **Two findings were about claims, not code** — a comment asserting an empty-string conclusion
  was handled when it was not, and an audit line reading `denylist acknowledged by .`, an
  acknowledgement by nobody. Both unreachable in practice; both fixed anyway, because a record
  that overstates what was proven is the specific failure #511 exists to end.
- **A test went vacuous during the fix** and was caught: it clobbered the executing script so
  the run died at exit 127 and the check never ran. It now fails loudly if the gate does not
  execute.

## Testing

80 assertions in `tests/test-merge-gate.sh`, 73/73 suites overall. Every new guard was verified
by removing it and confirming a test notices — necessary here because 8 of the new tests passed
*before* the implementation existed (an unknown flag made everything block, which is RED for the
wrong reason).

## This PR's own merge

It touches `scripts/merge-pr.sh`, and `--dual-certified` does not exist on `main` yet, so its
own merge is `--user-approved` one final time. Saying so rather than quietly doing it — that
framing honesty is the point of #511.


BaseInfinity:
Recorded from Fable's round-3 recheck, which explicitly said *"recording for honesty, not asking
for a change"* — and Codex certified the same tree, so it ships as-is rather than being fixed
after certification:

**P4 — a wrapped page carrying `"check_runs": null` would be misnamed in the error message.**
That payload is valid JSON, so `jq` parses it and the *filter* errors on `null[]`. The run
therefore fails closed, but the message says "could not parse the check-runs response as JSON",
which names the wrong cause. GitHub never emits it — `check_runs` is always an array — so the
path is unreachable.

Not fixed here deliberately. Changing code after both reviewers certified a specific tree would
mean shipping something neither actually reviewed, which is precisely the SHA-binding rule
(#437) this repo enforces on everything else. A one-word improvement is not worth breaking that
for. If it ever becomes reachable, the fix is `(.check_runs // [])[]`.

BaseInfinity:

**Reason:** Maintainer approved explicitly, 2026-08-08: 'yes approve one last time do the thing'. This is the LAST --user-approved of its kind: the PR builds --dual-certified, which does not exist on origin/main yet, so its own merge cannot use the path it ships. Both reviewers CERTIFIED HEAD at round 3 after three rounds of blind dual review (findings 7 -> 4 -> 0, none disputed); Codex verified by probe, Fable by mutation. CI validate green, no test deletions, no version bump — none of which were waived. Recorded honestly: the denylist tier and clearance-artifact checks were NOT verified, they were waived on this human decision.

- `scripts/merge-pr.sh (merge-evidence path)`
- ` clearance-artifact checks (.reviews/merge-clearance-517.json)`

Head: `67cdc032d0177926ac4ccd2e3135d037055911bc`

Issue #515, opened 2026-08-08
The unremarkable-PR path merges on an artifact the agent wrote itself — weaker evidence than every other tier


BaseInfinity:
Noticed while building `--dual-certified` (#511). Once that lands, `scripts/merge-pr.sh` has
four tiers of evidence, and they are **not** ordered by risk:

| PR touches | Evidence required |
|---|---|
| nothing on the denylist | a local JSON file the agent wrote itself |
| guidance (`SKILL.md`, wizard doc) | two visible, SHA-bound `YES` comments on the PR |
| merge-evidence chain | the same two comments, plus a gate byte-matching `origin/main` |
| `package.json` version | a human |

The first row is the loosest, and it is also the row that covers **most PRs** — ordinary
feature and fix work, which is exactly where a real bug ships. Its entire evidence is
`.reviews/merge-clearance-<PR>.json`: a gitignored local file, written by the same agent that
wrote the code, checked for `status`, `round >= 2`, a matching SHA, and a non-empty referenced
review file. Every one of those is a field the writer chooses.

The PR-comment path is not authenticated either — same `gh` token, as its own header says —
but it is strictly better in ways that matter against the stated threat model (an honest agent
that is degraded or under time pressure, not an adversary):

- it is **durable and visible**: a human reading the PR months later sees the evidence, and
  the visibility shape rules were hardened specifically so it cannot be collapsed or hidden
- it is **invalidated by any new push**, remotely, rather than by a local field the writer sets
- it requires **two distinct reviewer identities** and an explicit `YES` verdict at >=95 — the
  artifact requires neither a verdict nor a second reviewer

So the routine path asks for less evidence than the exceptional ones, and the asymmetry is
accidental rather than reasoned.

## The question to decide (not a foregone conclusion)

Should the two SHA-bound clearance comments become the standard evidence for **every** merge,
with the artifact demoted to a pointer at the dialogue? That would make the tiers monotonic:
same base evidence everywhere, escalating additions for riskier paths.

Argument against, which deserves a real hearing: the artifact path is cheap, and forcing two
posted comments onto every trivial PR is friction that will get routed around — and a gate
people route around is worse than a loose one. There may be a middle answer (require the
comments only above some blast radius) that is better than either extreme.

Route the decision through Fable before building anything.

## Explicitly not in scope

Server-side required checks. Still the only tamper-resistant answer, still parked.

[...]

PR #531, opened 2026-08-09
fix(autocompact): the wizard's own guidance was the thing that broke compaction

### LINKED ISSUE #520: Autocompact window guidance is wrong for 1M-context models (3 harness layers)
## What happened

Global `~/.claude/settings.json` carried `CLAUDE_CODE_AUTO_COMPACT_WINDOW: "350000"` (itself a replacement for an older `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE: "35"`). On Opus 5's 1M window this made the status line report two numbers that look contradictory but are both correct:

```
ctx:11% (890k left)        <- measured against the real 1M window
4% until auto-compact      <- measured against the artificial 350k window
```

Net effect: auto-compact fires with ~650k of usable context unused. Both knobs only ever move compaction *earlier* — neither raises real capacity. Removed from layer 1 on 2026-08-08.

## The three layers need three different answers

**Layer 1 — user global (`~/.claude/settings.json`)** — RESOLVED. Override removed; auto-compact now uses the model's real window.

**Layer 2 — this meta-repo (`.claude/settings.json`)** — OPEN. This repo has no `env` block at all today. Question: does a repo whose sessions routinely hold a Codex review, a Fable review, a full diff, and a 73-suite sweep want an explicit setting? Note the only available direction is *earlier* compaction, so the answer is plausibly "still nothing" — but it should be a decision, not an accident.

**Layer 3 — what the wizard ships (`CLAUDE_CODE_SDLC_WIZARD.md`)** — OPEN, and the real defect. Lines ~1044-1049 and ~1118 document both env vars and correctly warn "Do NOT set both." What they do NOT say is that on a 1M-context model you should set *neither* — the guidance reads as if picking one is the right move. A consumer following it on Opus 5 reproduces exactly the misconfiguration above. `skills/sdlc/SKILL.md` also pairs `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE=30` with a `claude-opus-4-6` pin, same trap.

## Proposed fix (layer 3)

Add an explicit "on a 1M-context model, set neither" line, and reframe the existing text so the vars are presented as *narrowing* tools for small-window pins (e.g. `opusplan` at 200K), never as tuning for large windows.

## Constraints

- Layer 3 is a shipped-doc change: TDD RED/GREEN + full cross-model round required.
- `skills/sdlc/SKILL.md` is at 19,356 / 20,000 bytes — any addition there needs an equal-or-greater trim.

BaseInfinity:
Closes #520. Process post-mortem in #532.

`CLAUDE_CODE_AUTO_COMPACT_WINDOW` and `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` **multiply**, so the wizard's recommended pairing produced a ~122k effective boundary on a 1M model. All three layers now say so — the wizard doc, the skill (and its cowork twin), and the instructions-loaded hook, which reports the real trigger and names where it read the setting from.

## The claim this PR originally got backwards

The first thirteen review rounds hardened the guards around a sentence that said a window under 200000 **disables** autocompact. It does not. It makes compaction fire **sooner**.

That came from reading `ZJu` — which arms the precomputed summary and does carry a `bIe = 200000` gate — and attributing that gate to the live trigger, which is `LXy` → `jUe` → `zJu` and has none. Two identifiers one letter-case apart. Verified in the v2.1.221 binary: `WINDOW=150000` → effective window 130000 → trigger `min(130000 × pct, 117000)`.

The shipped hook was telling every consumer session that autocompact was DISABLED and to raise the value — advice that moves the trigger *later* for someone whose config was already doing what they wanted.

**The guards were not wrong.** They pinned the false sentence faithfully, which is why the pins had to be flipped before the docs could be corrected. The retired wording is now a denylist entry on both surfaces.

## Tests that can actually fail

The guards went thirteen rounds, each defeating the last: vocabulary collision, deletion, semantic inversion, negating scope, then HTML comments / `<del>` / fences wrapping the claim, then bugs in the projection built to stop those.

Two design changes came out of it rather than patches:

- **Canonical-sentence pinning** — the load-bearing sentence is pinned verbatim, anchored to column 1. Token proximity cannot carry polarity, and a substring is not an assertion.
- **Rendered-text projection** (`tests/lib/mdfence.py`) — assertions match what a reader sees, not raw bytes. `--rendered` for prose claims, `--visible` for formula blocks. 49 selftest fixtures, each tagged with the round that produced it.

**Declared bounds are pinned as fixtures**, so a future change has to disagree deliberately: the scanner models block structure only.

## Verification

- Full suite 65/65; mdfence selftest 49/49; doc-consistency 129/129; hooks 170/170
- Skill twins byte-identical (19,638 bytes)
- Mutation harness on every new assertion — attacks CAUGHT, false-positive controls GREEN
- Codex `high`: CERTIFIED at round 13 **on the guards** — that certification did not cover claim truth, and is superseded


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Codex GPT-5.6 Sol at high effort, thirteen review rounds on review_id 520-autocompact-guidance, CERTIFIED at round 13 with no P0/P1/P2 findings. Confidence and verdict requested separately and quoted verbatim; the number is the reviewer's, not mine.

Residual named by the reviewer: environment-specific behaviour not exercised by the reviewed test suite.

```json
{"verdict": "YES", "confidence": 97, "reviewer": "codex-gpt-5.6-sol", "sha": "b05dc64035d5e0c13899d1fe6fc05b916f63fe55"}
```

BaseInfinity:

**Reason:** maintainer authorized in session; Codex CERTIFIED + Fable fresh-eyes, CI green on d5fff3d; hooks/ denylist hit is the self-produced-evidence circularity tracked by #515/#540

- `hooks/instructions-loaded-check.sh (merge-evidence path)`
- ` clearance-artifact checks (.reviews/merge-clearance-531.json)`

Head: `d5fff3de638733577310ef5a362708fe352b2cdf`

Issue #483, opened 2026-08-02
Research: at what context size does Opus 5 quality actually degrade — and should the wizard ship a threshold?


BaseInfinity:
## Question

Maintainer, 2026-08-02, during a session that reached **~687K tokens**: *"do we have an 'audit your token and autocompact time'? I hear past 250-350K is a nightmare."*

We ship autocompact *documentation* but no *recommendation* for Opus 5, and no measurement behind any number. Decide whether a threshold should be recommended, and if so what it is.

## What already exists

| Thing | What it does | Gap |
|---|---|---|
| `hooks/token-spike-check.sh` | SessionStart hook; warns if last session's burn deviates >2σ above rolling median (ROADMAP #220) | Detects **cost** anomalies, not **quality** degradation. Says nothing about context size. |
| `hooks/precompact-seam-check.sh` | Guards the compaction seam | Operational, not advisory |
| Wizard doc autocompact section | Documents `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` and `CLAUDE_CODE_AUTO_COMPACT_WINDOW`, with a measured Sonnet 5 figure (~967K / 96.7%) | **No Opus 5 threshold.** The section explicitly notes the override only causes earlier compaction where CC compacts proactively, and no proactive Opus 5 threshold is documented. |

So: we can detect a cost spike and we can tell a consumer *how* to set a threshold. We cannot tell them *what to set it to*, or whether to bother.

## What is claimed vs what is known

- **Community claim:** quality degrades badly past 250–350K. Source is hearsay as recorded here — unverified, no methodology, model unspecified.
- **Repo memory** pairs `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE=30` with Opus 4.6's 1M context, and warns *not* to set it for `opusplan` (200K, too aggressive). That is a config note, not a quality measurement.
- **Anthropic documents** Sonnet 5's tuned proactive default (~96.7%), which implies they do NOT believe quality collapses at 30% of window — at least for Sonnet.
- **Live data point:** this session ran to ~687K on Opus 5 (1M) and produced its most valuable findings late — a shipped hook proven silently dead, a guard proven vacuous. That is one uncontrolled anecdote, and it cuts *against* the 250–350K claim, which is exactly why it needs measuring rather than asserting.

## What to determine

[...]

PR #549, opened 2026-08-10
fix(audit): delete the unmeasured 20,000-byte ceiling gate (#489)

### LINKED ISSUE #489: skills/sdlc/SKILL.md is at its 20,000-byte ceiling — every edit now starts by degrading something else

BaseInfinity:
Closes #489.

## What

Deletes the CI gate that turned the session-load audit's TRIM flag into a build failure for `skills/sdlc/SKILL.md`. The audit tool itself is untouched — it still inventories every session-loaded asset and still raises TRIM rows.

## Why the number could not be enforced

`scripts/audit-session-load.sh` computes the threshold as `THRESHOLD_TOKENS × 4` — 5,000 (a round number) times 4 (a chars-per-token rule of thumb from a different vendor) — and `SDLC_AUDIT_THRESHOLD_TOKENS` moves it at will. Under the CLAIM RULE we do not enforce a number we never measured. #483 is where a measured ceiling would come from; it is still open.

## Why the cost was real

`skills/sdlc/SKILL.md` sat at **19,997 of 20,000 bytes**. Every subsequent edit had to buy its own bytes by degrading another instruction in the same file, and that trading shipped three defects in a single session — including invalid JSON and a dropped instruction value. A budget that forces a silent trade on every edit is a circling generator, which is the exact failure this milestone exists to end.

## The gate was never a coherent bloat policy

Found by Fable during review, verified independently by running the audit at base. The gate filtered `type == "skill"` only. Both files the audit actually flags today are ones it never touched:

```
40795  10198  hook          TRIM  hooks/instructions-loaded-check.sh
25711   6427  instructions  TRIM  SDLC.md
```

It failed the build over the smallest of the three oversized files while twice-larger shipped files sailed past.

## Coverage that was nearly lost

GPT-5.6 filed a P1 in round 1: the deleted test was the only place a real `json.loads` ran over the audit's REAL multi-entry output. The surviving JSON test greps substrings; the other parser test uses a single-entry fixture. Dropping the inter-entry commas left all nine survivors green.

Replaced with `test_audit_json_is_parseable_on_real_repo`, which asserts validity, `>= 2` entries, per-entry `path`/`type`/`flag`, and `trim_candidate_count` — and asserts **nothing** about sizes or flags, so it cannot become the ceiling gate by the back door. It passes today with `trim_candidate_count = 2`, which is the live proof of that.

Fable's audit of the swap found the replacement is a strict superset of the old test's JSON coverage: the old test *silently passed* a non-list `entries`, a missing per-entry key, and a missing `trim_candidate_count`, because the resulting `TypeError`/`KeyError` was swallowed by `2>/dev/null`. The only behaviour lost is the gate itself.

## Accepted, documented risk

Nothing automated runs the audit in report mode — the only invoker outside the test file is CI running the test suite. TRIM rows now surface **on demand**, when someone runs `scripts/audit-session-load.sh`. Until #483 produces a measured ceiling there is no automated bloat backstop for shipped skills. That is the deliberate trade, not an oversight.

## Verification

- RED/GREEN proven both directions: 20 bytes over the threshold → `Failed: 1` at base, `Failed: 0` on this branch, with the audit still printing the TRIM row.
- 65/65 non-E2E CI suites green locally; shellcheck clean (only the pre-existing SC2295, identical on base).
- `tests/test-cowork-drift.sh` byte-parity untouched: 31/31.
- `scripts/audit-session-load.sh` is comment-only — `--json` byte-identical to base at the default threshold and under `SDLC_AUDIT_THRESHOLD_TOKENS` override.

## Review

Two reviewers, two rounds each, independently:

- **GPT-5.6 Sol `high`** — round 1 NOT CERTIFIED (one P1, accepted and fixed); round 2 **CERTIFIED**, confidence 99, after running its own five JSON malformations, all caught.
- **Fable `xhigh`** — **CLEAR**, zero unresolved findings, after applying the diff to a scratchpad copy of base, running its own mutation RED proof, and fact-checking every load-bearing claim in the new comments.

### DIFF
--- a/scripts/audit-session-load.sh
+++ b/scripts/audit-session-load.sh
@@ -9,6 +9,14 @@
 # a mechanical inventory of every file/output that hits a session, so trim
 # candidates surface before they bloat the context budget.
 #
+# #489: THIS TOOL IS ADVISORY. The TRIM threshold is not a measured ceiling —
+# it is 5,000 (a round number) times 4 (a chars-per-token rule of thumb from a
+# different vendor), and SDLC_AUDIT_THRESHOLD_TOKENS can move it at will. A
+# test once turned that flag into a CI failure, which made an unmeasured number
+# a hard contract on a shipped file and forced every later edit to buy its
+# bytes by degrading something else. That gate is deleted. No measured ceiling
+# exists today; #483 is where one would come from.
+#
 # Usage:
 #   scripts/audit-session-load.sh           # human-readable table
 #   scripts/audit-session-load.sh --json    # machine-readable
--- a/tests/test-audit-session-load.sh
+++ b/tests/test-audit-session-load.sh
@@ -143,33 +143,68 @@ test_audit_ranks_by_size_descending() {
     fi
 }
 
-# Eat-our-own-dogfood: the wizard's own SKILL.md files must come in below the
-# bloat threshold the audit warns about. Phase 2 follow-up to PR #272 — the
-# audit tool flagged 2 of our 4 SKILL.md files (sdlc 12,427 tokens; update
-# 8,555 tokens) on the day we shipped it. Acting on the tool's findings closes
-# the Prove-It loop: a tool that surfaces real issues whose owner ignores them
-# is just a louder lint warning.
-test_wizard_own_skills_below_threshold() {
+# #489: the build gate that used to live here is deleted.
+#
+# It failed CI whenever a SKILL.md crossed the audit's TRIM threshold, which
+# made 20,000 bytes a hard contract on a shipped file. That number was never
+# measured: `scripts/audit-session-load.sh` computes it as THRESHOLD_TOKENS
+# (5,000, a round number) times 4 (a chars-per-token rule of thumb from a
+# different vendor), and the whole thing is tunable by an environment variable.
+# Under the CLAIM RULE we cannot enforce a number we never measured.
+#
+# The cost was real and compounding. With sdlc/SKILL.md sitting at 19,997 of
+# 20,000, every subsequent edit had to buy its own bytes by degrading another
+# instruction in the same file, and that trading shipped three defects in a
+# single session — including invalid JSON and a dropped instruction value.
+# A budget that forces a silent trade on every edit is a circling generator.
+#
+# The audit itself is untouched: it still reports sizes and still raises TRIM,
+# because observability was always its job. Only build-red is gone. If a real
+# ceiling is ever measured, it belongs here — see #483, which is where such a
+# measurement would come from.
+#
+# Still enforced elsewhere, deliberately: tests/test-cowork-drift.sh keeps
+# skills/sdlc/SKILL.md and cowork/skills/sdlc/SKILL.md byte-identical.
+#
+# One thing the deleted test was doing incidentally, kept below: it was the
+# only place that ran a real json.loads over the audit's REAL multi-entry
+# output. Cross-model review caught that removing the gate would have taken
+# that coverage with it — the other JSON test greps for substrings, so
+# stripping the commas between entries left all nine survivors green while
+# `python3 -m json.tool` rejected the output.
+
+# Parses the audit's output for THIS repo, which is the only fixture with
+# enough entries for a malformed separator to show up. Asserts validity and
+# shape only — never sizes, never flags. That distinction is the point: the
+# tool's contract is that it emits parseable JSON, not that any file in this
+# repo is under some particular number.
+test_audit_json_is_parseable_on_real_repo() {
     local repo_root="$SCRIPT_DIR/.."
-    local output flagged
+    local output verdict
     output=$(SDLC_AUDIT_ROOT="$repo_root" "$AUDIT" --json 2>&1)
-    flagged=$(printf '%s' "$output" | python3 -c '
+    verdict=$(printf '%s' "$output" | python3 -c '
 import json, sys
 try:
     data = json.loads(sys.stdin.read())
-except json.JSONDecodeError:
-    print("INVALID_JSON")
+except json.JSONDecodeError as e:
+    print("INVALID_JSON: %s" % e)
     sys.exit(0)
-flagged = [e["path"] for e in data.get("entries", [])
-           if e.get("type") == "skill" and e.get("flag") == "TRIM"]
-print("\n".join(flagged))
+entries = data.get("entries")
+if not isinstance(entries, list):
+    print("NO_ENTRIES_LIST")
+elif len(entries) < 2:
+    print("TOO_FEW_ENTRIES: %d" % len(entries))
+elif any(k not in e for e in entries for k in ("path", "type", "flag")):
+    print("ENTRY_MISSING_KEYS")
+elif "trim_candidate_count" not in data:
+    print("NO_TRIM_COUNT")
+else:
+    print("OK")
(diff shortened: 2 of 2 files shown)


BaseInfinity:
## Cross-model review record

`.reviews/` is gitignored, so this comment is the durable record.

### GPT-5.6 Sol `high` — 2 rounds

**Round 1 — NOT CERTIFIED**, confidence 99. One P1:

> Multi-entry JSON validity coverage was unintentionally removed. The deleted dogfood test used `json.loads` on real, multi-entry audit output. The retained JSON test only greps strings; the remaining parser test uses one entry.

It proved this by removing inter-entry commas in an isolated worktree: all nine retained tests passed while `python3 -m json.tool` failed. Accepted, not disputed. Fixed with `test_audit_json_is_parseable_on_real_repo`.

**Round 2 — CERTIFIED**, confidence 99, zero findings. It was asked to verify using its *own* malformations rather than mine:

> All five independent JSON malformations were caught. At 20,016 bytes, all 10 audit tests passed and the audit retained its TRIM row. Default and overridden JSON output matched base byte-for-byte. The other nine tests and cowork parity enforcement are unchanged. All 65 non-E2E suites passed.

### Fable `xhigh` — 2 rounds

**Round 1 — CLEAR**, zero unresolved findings. Method: extracted base into a scratchpad copy, applied the diff with `patch(1)`, ran the audit suite (10/10) and cowork drift (31/31), ran its own mutation RED proof, and fact-checked every load-bearing claim in the new comments.

Two things it found that the Codex rounds and I had both missed:

1. **The gate was never a coherent bloat policy.** It filtered `type == "skill"` only. Both files the audit actually flags today are ones it never touched — `hooks/instructions-loaded-check.sh` at 40,795 chars and `SDLC.md` at 25,711. It failed the build over the smallest of the three oversized files. Verified independently by running the audit at base.

2. **The replacement test is a strict superset, not an equivalent.** The old test *silently passed* a non-list `entries`, a missing per-entry key, and a missing `trim_candidate_count` — the resulting `TypeError`/`KeyError` was swallowed by `2>/dev/null` into an empty string, which hit its pass branch. The only behaviour lost in the swap is the gate itself.

**Round 2 — CLEAR**, on the dialogue round about where its observability caveat should live. It ruled the in-code comment is *under-informative but not misleading* — "Only build-red is gone" is true and the correct inference is derivable — so the caveat belongs in the PR body rather than in an amended diff, and stated plainly it was not choosing that because it was the cheap option:

> If the comment had said "TRIM candidates still surface in CI," I would say (a) and eat the round.

It also confirmed the one nit it raised (an empty `got:` string when a non-dict entry raises) should *not* be fixed here, because there is no path from the swallowed exception to a false green — the pass branch requires `verdict = "OK"` exactly, so the defect is confined to the diagnostic string on an already-red test.

### SHA binding

Both round-2 clearances were issued against `435708a` because the work was uncommitted at review time. The commit `f4f2440` was verified byte-identical to the reviewed snapshot before pushing. Fable's clearance self-transfers on that condition by its own terms; GPT-5.6's was rebound in a round 3 that re-verified the diff and fact-checked the commit message's claims.

BaseInfinity:
### Round 3 — SHA rebind, GPT-5.6 Sol `high`

Rounds 1 and 2 were issued against `435708a`, the base commit, because the work was still uncommitted at review time. `scripts/merge-pr.sh` compares the clearance SHA to the PR head, so a clearance naming the base cannot authorize this merge. Round 3 rebinds it — and re-verifies rather than rubber-stamping the transfer.

It was asked to confirm the diff independently *and* to fact-check every claim in the commit message, on the grounds that a false claim in a permanent commit message is exactly what the CLAIM RULE exists to catch:

> No differences found. Diff contains exactly two files and the reviewed hunks only. Working tree is clean. Range contains exactly one commit. Threshold, override, 19,997-byte size, skill-only deleted gate, current TRIM entries/sizes, preserved TRIM behavior, and report-mode claim all verified. Commit-message claims are supported.

**CERTIFIED**, confidence 99, bound to `f4f2440a9299ed56b6377ecbdedea09fb1eb7584`.

Fable's round-2 clearance was issued conditional on the commit matching the reviewed snapshot byte-for-byte. `git diff 435708a f4f2440` was verified identical to that snapshot before the push, so it transfers by its own stated terms without a further round.

Issue #551, opened 2026-08-10
Parked: a doc guard that no doc INSTRUCTS 'sudo npm install -g' (dropped from #476 on process budget)


BaseInfinity:
Parked from #476. The documentation shipped; this guard did not. Reopen only if a real regression happens — not on principle.

## What it was

A test asserting that no consumer-facing doc *instructs* `sudo npm install -g @anthropic-ai/claude-code`. The hard part is that it cannot be "the string must not appear", because the warning has to quote the command it warns about. So it had to distinguish a warning from an instruction.

## Why it was dropped

Six review rounds, two reviewers. The rule was rewritten three times and every rewrite was broken by a sentence neither reviewer nor I had thought of:

| Rule | Broken by |
|---|---|
| negation anywhere on the line | `Do not use the native installer; run sudo npm install -g ... instead.` |
| negation in the same clause | `Do not forget to run sudo npm install -g ...` (litotes) |
| negation bound through an allow-gap | `Do not avoid using sudo npm ...` (regex restarts at the inner negator) |
| + counting negator tokens | `You should not avoid using sudo npm ...` (`should not` was never a counted token) |
| anchored at the clause's first token | *(rule held)* — then the **needle** started failing: `sudo -E npm`, then `sudo -u root npm` |

The rule converged. The needle did not, and cannot: `as root, run npm install -g ...` names the same footgun with no `sudo` token in it at all. Widening is unbounded by construction.

Dropped under the PROCESS BUDGET rule from #531. A test may not cost more review rounds than the change it guards — the issue asked for ~20 documentation lines.

## What shipped instead

`test_wizard_prereqs_recommend_native_claude_install`, which is section-scoped to the wizard's `## Prerequisites` and asserts the install command is present, is described as recommended, and is not negated. It was never contested after round 1. Cross-model review covers what a regex cannot.

## What is worth keeping if this is ever revived

The parked implementation had one genuinely good property: a 27-case corpus in which **every case was a sentence a reviewer actually used to break the guard**, proven bidirectionally live (make the rule always-true → red; always-false → red the other way; no dead entries). That corpus is the valuable artifact, not the regex.

Two findings from the same cycle that outlived the guard and are worth remembering:

[...]

Issue #550, opened 2026-08-10
Local 'all suites green' claims are unreliable: a test script eats stdin in a while-read loop


BaseInfinity:
## What happened

Running the non-E2E CI suites locally with the obvious harness:

```bash
while IFS= read -r t; do "$t" >> log 2>&1 || echo "FAILED: $t"; done < suites.txt
```

reported **30 suites passed, 0 failed**. There are **65** non-E2E suites in `ci.yml`, and all 65 exist and are executable. No suite was reported skipped, because none was: the loop simply never reached them.

Cause: at least one test script reads from stdin, and stdin inside the loop is `suites.txt`. The script consumes the rest of the list. The loop then ends normally and reports success.

## Why it matters

The failure is silent and reports **green**. Nothing in the output distinguishes "65 suites passed" from "30 suites passed and the harness ate the other 35" except counting — and the count is what a person is least likely to check when the line says `FAILED: 0`.

Any local "all suites pass" claim in this repo made with a `while read` harness is suspect, including ones already recorded in PR bodies and review handoffs. CI is unaffected: `ci.yml` invokes each suite as its own step.

## Fix

Close stdin per suite:

```bash
for t in $(grep -v e2e suites.txt); do "$t" < /dev/null >> log 2>&1 || echo "FAILED: $t"; done
```

With that, the same list gives 65/65.

## Suggested follow-ups

- Identify which suite(s) actually read stdin — they may have the same latent problem under other runners.
- Consider having the harness assert the number of suites it ran against the number in `ci.yml`, so a shortfall is loud rather than green. This is the same class as the guards this repo already has: a check that cannot fail is worse than no check.

## Related

Found during #476, in a session that also shipped a doc guard which matched nothing and passed vacuously (`[[:space:]]` handed to a Python regex), and where a reviewer's own verification matrix was vacuous for a third reason (`r""` is falsy in Python). Three independent dead checks in one cycle — the pattern is worth a rule, not just three fixes.

BaseInfinity:
Same family, spotted while auditing a green CI run (PR #552, run 31365065132):

```
validate  Run self-update mechanism tests  ./tests/test-self-update.sh: line 191: echo: write error: Broken pipe
validate  Run self-update mechanism tests  ./tests/test-self-update.sh: line 215: echo: write error: Broken pipe
validate  Run self-update mechanism tests  ./tests/test-self-update.sh: line 402: echo: write error: Broken pipe
```

[...]

PR #552, opened 2026-08-10
docs(install): recommend the native Claude Code installer, warn off sudo-npm (#476)

### LINKED ISSUE #476: Recommend native install for Claude Code CLI (avoid sudo-npm footgun)

BaseInfinity:
Closes #476.

## What

The wizard documented how to install **itself** and never how to install Claude Code, the CLI it runs on. Adds a Prerequisites block naming the native installer as the recommended path, one README sentence, and `which -a claude` as the conflict check.

## The footgun this steers off

Observed on a real machine (#476): `sudo npm install -g @anthropic-ai/claude-code` left the global module directory root-owned. From there `claude update` refused with "Insufficient permissions to install update"; `npm uninstall -g` failed with `EACCES`, because a normal user cannot rename a root-owned directory; and switching to the native install left *two* `claude` binaries on `PATH`, resolved by `PATH` order.

## `which -a claude`, not `claude doctor`

Review ran `claude doctor` 2.1.221 against a decoy binary placed earlier on `PATH`. It reported **no conflict** — precisely the PATH-shadowing case the prose describes. The remediation I first wrote would have masked the exact failure it existed to catch. The official docs' own procedure is `which -a claude`; doctor stays for install health with its blind spot stated.

## Claims scoped to what is actually documented

Every factual claim was checked against Anthropic's [setup docs](https://code.claude.com/docs/en/setup) by both reviewers:

- sudo **can** leave the directory root-owned, not does — the docs say "can lead to permission issues"
- the compounding failures are attributed as observed on one machine, not asserted as guaranteed
- a non-sudo npm install stops auto-updating "with only a one-time notice at startup", not *silently* — the docs state a notice is shown

## The test

`test_wizard_prereqs_recommend_native_claude_install` is scoped to the `## Prerequisites` section, not the file. Asserting a string appears somewhere in 4,900 lines is the #493 failure mode, and this doc already names the installer in an unrelated aside about version pinning.

It asserts three things, each RED-proven: the install command is present; it is described as recommended; and no negation reaches it. That third one exists because an earlier version asserted only the URL — flipping the prose to "Do not use the native installer" while keeping the URL still passed. The guard proved a string, not a stance.

## What was dropped, and why

A companion guard asserting no doc *instructs* sudo-npm is parked in **#551**. It cannot be "the string must not appear", since the warning has to quote the command it warns about, so it had to tell a warning from an instruction. Six rounds, two reviewers, three rewrites of the rule — each broken by a sentence nobody had thought of, ending with the rule converging but its **needle** not converging: `sudo -E npm`, then `sudo -u root npm`, and `as root, run npm install -g` names the same footgun with no `sudo` token at all.

Dropped under the PROCESS BUDGET rule from #531 — a test may not cost more review rounds than the change it guards, and this issue asked for ~20 documentation lines. Both reviewers agree the documentation ships without it: GPT-5.6 chose the deletion explicitly and certified conditional on it.

Two findings from that work outlived the guard and are recorded in #551 and #550:

1. The guard was once **completely dead** — `sudo[[:space:]]+npm` handed to a *Python* regex, where `[[:space:]]` is the character set `[ : s p a c e` and matches no space. It matched nothing, passed vacuously on every input, and the suite was fully green.
2. A local suite harness silently ran **30 of 65** suites and reported success, because a test script consumes stdin and ate the rest of a `while read` list (#550).

A check that cannot fail is worse than no check, and the only thing that reliably found these was mutating the shipped artifact and watching it go red.

## Verification

- All three assertions of the surviving test proven RED individually, then GREEN.
- 65/65 non-E2E CI suites (with stdin closed per suite — see #550).
- shellcheck identical to base (43 findings, none in the added block).

## Review

- **GPT-5.6 Sol `high`** — 6 rounds. Findings: inverted-recommendation acceptance, semantic bypasses, prose overstatement, a false comment about npm not shipping README (`npm pack --dry-run` proves it does), and two guard bypasses. **CERTIFIED** conditional on dropping the parked guard.
- **Fable `xhigh`** — 4 rounds. Findings: the `claude doctor` blind spot, first-mention-per-line blindness, a comment overclaiming soundness, needle evasion. **CLEAR**, zero unresolved.

### DIFF
--- a/CLAUDE_CODE_SDLC_WIZARD.md
+++ b/CLAUDE_CODE_SDLC_WIZARD.md
@@ -246,6 +246,16 @@ When Anthropic provides official plugins or tools that handle something:
 | **Claude Code v2.1.69+** | Required for InstructionsLoaded hook, skill directory variable, and Tasks system |
 | **Git repository** | Files should be committed for team sharing |
 
+**Installing Claude Code itself.** Use the native installer — it is what [the official setup docs](https://code.claude.com/docs/en/setup) label *Recommended*, and it auto-updates in the background with no `claude update` step to remember:
+
+```bash
+curl -fsSL https://claude.ai/install.sh | bash
+```
+
+**Never** install it with `sudo npm install -g @anthropic-ai/claude-code` — the official docs warn against that in those words, because sudo can leave the global module directory root-owned. When it does, the damage compounds, as observed on a real machine (#476): `claude update` refused with "Insufficient permissions to install update"; `npm uninstall -g` then failed with `EACCES`, because a normal user cannot rename a root-owned directory; and switching to the native install left *two* `claude` binaries on `PATH` at once, resolved by `PATH` order — so a reordering downgrades you to the stale pinned version. An npm install *without* sudo is not dangerous, just worse: if the global npm directory is not writable it stops auto-updating, with only a one-time notice at startup.
+
+Run `which -a claude` to check for conflicting installations — that is the official docs' own procedure, and it is the one that catches a stale binary shadowing a newer one by `PATH` order. Use `claude doctor` for install health, but do not rely on it for this: when run with a decoy binary placed earlier on `PATH`, it reported no conflict.
+
 **Blank repos (no CLAUDE.md, no code):** The wizard works on empty repos. Run `npx -y agentic-sdlc-wizard@latest init` — it installs hooks, skills, and the wizard doc. (The `@latest` pin guards against stale npx caches per #358.) On first session, the hooks detect missing SDLC files and redirect to `/claude-setup-wizard`, which generates CLAUDE.md, SDLC.md, TESTING.md, and ARCHITECTURE.md interactively. You do NOT need to run Claude's built-in `/init` first — the setup wizard handles everything.
 
 ---
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A **self-evolving Software Development Life Cycle (SDLC) enforcement system for 
 
 ## Install
 
-**Requires [Claude Code](https://docs.anthropic.com/en/docs/claude-code/overview)** (Anthropic's CLI for Claude).
+**Requires [Claude Code](https://docs.anthropic.com/en/docs/claude-code/overview)** (Anthropic's CLI for Claude). Install it with the native installer — `curl -fsSL https://claude.ai/install.sh | bash` — which the [official setup docs](https://code.claude.com/docs/en/setup) label *Recommended* and which keeps it auto-updating in the background. **Never** use `sudo npm install -g @anthropic-ai/claude-code`: sudo can leave the global module directory root-owned, which then breaks `claude update` *and* `npm uninstall -g`. Check for conflicting installs with `which -a claude`.
 
 Run from your terminal or from inside Claude Code (`!` prefix):
 ```bash
--- a/tests/test-doc-consistency.sh
+++ b/tests/test-doc-consistency.sh
@@ -3668,6 +3668,66 @@ $out"
 }
 test_guard_selftests_actually_run
 
+# #476: the wizard documented how to install ITSELF and never how to install
+# Claude Code, the thing it runs on. A real machine hit the footgun the official
+# setup docs warn about in those words: `sudo npm install -g` left the global
+# module dir root-owned, which killed `claude update` ("Insufficient
+# permissions"), then killed `npm uninstall -g` (EACCES on a root-owned rename),
+# and finally left two `claude` binaries on PATH at once — resolved by PATH
+# order, so a reordering would have downgraded the user with no warning.
+#
+# Scoped to the Prerequisites SECTION on purpose. "the installer is named
+# somewhere in a 4,900-line file" is the #493 failure mode — it passes on a doc
+# that mentions the installer in an unrelated aside about version pinning, which
+# this doc already contains at the version-test benchmark procedure. The claim
+# under test is narrower and is the one that matters: a reader standing at the
+# install moment is told which installer to use.
+#
+# The recommendation half exists because round 1 asserted only that the URL
+# appeared. Rewriting the prose to "Do not use the native installer" while
+# keeping the URL still passed — the guard proved a string, not a stance.
+test_wizard_prereqs_recommend_native_claude_install() {
+    local DOC="$REPO_ROOT/CLAUDE_CODE_SDLC_WIZARD.md"
+    if [ ! -f "$DOC" ]; then fail "CLAUDE_CODE_SDLC_WIZARD.md not found"; return; fi
+    local section tmp negated
+    section=$(awk '/^## Prerequisites$/ { f = 1; next } /^## / { f = 0 } f' "$DOC")
+    if [ -z "$section" ]; then
+        fail "#476: no '## Prerequisites' section in the wizard doc to carry install guidance"
+        return
+    fi
+    tmp=$(mktemp "${TMPDIR:-/tmp}/prereq-XXXXXX")
+    printf '%s\n' "$section" > "$tmp"
+
+    if ! grep -qE 'claude\.ai/install\.sh' "$tmp"; then
+        fail "#476: wizard Prerequisites gives no Claude Code install command, so a reader at the install moment is never steered off sudo-npm"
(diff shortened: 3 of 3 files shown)


BaseInfinity:
### Cross-model review record

`.reviews/` is gitignored, so this comment is the durable record.

**GPT-5.6 Sol `high` — 7 rounds.** Findings across the dialogue: the Prerequisites test accepted an inverted recommendation (URL present, prose flipped to "Do not use the native installer" — still green); two semantic bypasses of the sudo guard; prose overstating what Anthropic documents; a false comment claiming npm does not ship README (`npm pack --dry-run` proves it does); and two further guard bypasses. On being asked for a terminal decision it chose to **drop the guard** — "the semantic space is unbounded, and the repo's positive-anchor test is sufficient regression coverage for the requested documentation stance" — and certified conditional on that deletion. Rebound to the PR head after commit:

> No findings. Exact three-file diff; deleted guard, helper, and `negbind.py` are absent. One commit; branch and worktree clean; PR head matches. Both required mutations went RED; restored baseline is 130/0. Non-E2E suites: 65/65. ShellCheck unchanged, zero findings in the added block. Permanent claims verified against Anthropic's setup and troubleshooting docs, live `claude doctor` 2.1.221 decoy testing, `npm pack --dry-run`, issue #476, and the recorded reviewer clearances.

**CERTIFIED**, confidence 99, bound to `2094700`.

**Fable `xhigh` — 4 rounds, CLEAR, zero unresolved.** Its findings drove most of the prose:

- **`claude doctor` does not detect the failure the doc told readers to use it for.** It ran doctor 2.1.221 with a decoy `claude` earlier on `PATH` and got a clean report — precisely the PATH-shadowing case the prose describes. The remediation as first written masked the exact failure it existed to catch. `which -a claude` is the docs' own procedure and is what shipped.
- "silently stops auto-updating" overstated; the docs state a one-time startup notice is shown.
- A comment overclaiming that a guard was sound.

Fable's clearance was issued against the superset diff that still carried the sudo guard. The shipped diff is a strict subset of it — the guard removed — and GPT-5.6 chose and certified that removal explicitly, so no leg of the evidence covers less than what shipped.

### On the dropped guard

Parked in #551 with its 27-case corpus. Both reviewers converged on dropping it: Fable established that needle coverage is unbounded in principle (`as root, run npm install -g` names the same footgun with no `sudo` token), and GPT-5.6 chose deletion when asked to decide. It was dropped under the **PROCESS BUDGET** rule shipped in #531 — a test may not cost more review rounds than the change it guards.

Recorded plainly because it is the useful part: this PR's review went six rounds on a ~20-line documentation change, which is the circling behaviour milestone v1.97.0 exists to eliminate. The rule that ends it had already merged and was not applied. Rounds 1–4 bought real defects — a guard that matched nothing and passed vacuously, a false `claude doctor` claim, a false npm-README claim. Rounds 5–6 bought spellings of `sudo`. The budget should have cut between those.

BaseInfinity:
### Cross-model clearance — GPT-5.6 Sol `high`, 7 rounds

Rebound to the PR head after commit. It verified the three-file diff, confirmed the dropped guard and helper are absent, re-ran both RED mutations on the surviving test, and fact-checked the permanent commit-message and PR-body claims against Anthropic's setup and troubleshooting docs, live `claude doctor` 2.1.221 decoy testing, and `npm pack --dry-run`.

```json
{"reviewer":"gpt-5.6-sol","verdict":"YES","confidence":99,"sha":"209470042c5f3339588d9031e5906a66d0c33be7"}
```

BaseInfinity:
### Cross-model clearance — Fable `xhigh`, 5 rounds

Reviewed the diff that actually ships, not the superset it cleared earlier — the guard was dropped after its round 4, so this is a different artifact.

Asked specifically whether the surviving test's third assertion, now a single grep where a module used to be, is a weaker check wearing the old one's clothes. It compared the two directly on every divergence it could construct:

> The grep is identical-or-more-paranoid everywhere I probed, and it closes the only attack ever made against this test. On litotes ("Don't forget to use the native installer") the grep goes RED where the old rule was green. On the comma-smuggled inversion the grep is false-green — and I verified the dropped negbind rule was false-green on the *same* sentence. Shared hole, adversarial class, both disclosed.

All three assertions mutation-proven independently. Docs verified by blob identity (`4b32df3`, `855f7a8`) against the tree whose every prose fact it had already checked against Anthropic's setup and troubleshooting docs. CI `validate` confirmed via the GitHub API on this exact head SHA rather than on assertion.

On the drop itself: *"I endorse it. It is my own round-3 position carried to its conclusion — needle-widening was tripwire extension, never soundness, and the spelling surface is unbounded."*

```json
{"reviewer":"fable","verdict":"YES","confidence":97,"sha":"209470042c5f3339588d9031e5906a66d0c33be7"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Reviewer: GPT-5.6 Sol `high`, 7 rounds. Rebound to the PR head after commit; verified the three-file diff, confirmed the dropped guard and helper are absent, re-ran both RED mutations on the surviving test, and fact-checked the permanent commit-message and PR-body claims against Anthropic's setup and troubleshooting docs, live `claude doctor` 2.1.221 decoy testing, and `npm pack --dry-run`.

```json
{"reviewer":"gpt-5.6-sol","verdict":"YES","confidence":99,"sha":"209470042c5f3339588d9031e5906a66d0c33be7"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Reviewer: Fable `xhigh`, 5 rounds. Reviewed the diff that actually ships rather than the superset it cleared earlier. Compared the surviving test's grep against the machinery it replaced and found it identical-or-more-paranoid on every divergence probed; all three assertions mutation-proven independently; docs verified by blob identity against the tree whose prose it had already fact-checked; CI `validate` confirmed via the GitHub API on this head SHA.

```json
{"reviewer":"fable","verdict":"YES","confidence":97,"sha":"209470042c5f3339588d9031e5906a66d0c33be7"}
```

Issue #557, opened 2026-08-10
Ship the three /sdlc rules that stop a #539-shaped review loop before round 1


BaseInfinity:
## Scope card
- **One issue:** ship the three /sdlc lines that stop a review loop like #539's before round 1.
- **Acceptance:** the three rules below present in `skills/sdlc/SKILL.md` and mirrored in `CLAUDE_CODE_SDLC_WIZARD.md`, cowork copy byte-identical.
- **Allowed paths:** `skills/sdlc/SKILL.md`, `cowork/skills/sdlc/SKILL.md`, `CLAUDE_CODE_SDLC_WIZARD.md`
- **Exclusions:** **NO regex guards greping for these new sentences.** That would violate rule 2 the moment it ships — the lesson of this very issue. No new counter: #530 deleted a duplicate round counter for contradicting the existing one, and the guard-cost rule already carries this distinction.
- **Risk tier:** low
- **Estimated diff:** ~10 lines across three files

## Where this came from

PR for #539 + #538: a 23-line doc change, correct at round 1 and never changed again, carried a 197-line grep guard over prose. **Four review rounds, seven blocking findings, all seven in the test, zero in the deliverable.** Rounds 2–4 were the same defect three times: a text-presence check cannot detect polarity. Final defeat was Markdown strikethrough — `~~the count never resets~~` keeps the searched string byte-identical while the live sentence teaches the opposite; the suite stayed 137/137 green. The guard was deleted under a pre-registered tripwire.

Both reviewers were asked independently how this could have been short-circuited, and reconciled.

## The three lines

**1. Replace the TDD RED line** — *"watch EACH assertion fail, not the suite"* is not sufficient. Every RED proof this session was red-by-deletion, the maximal mutation. Replacement: enumerate the check's promised observation set; on a copy of the **live deliverable**, break each member separately and require its named failure; the runner fails unless every expected suite reports. **And require the NEAREST WRONG form to fail, not merely deletion** — a branch satisfied by pre-existing text is invisible to deletion-mutation. If the set cannot be enumerated, do not write the guard.

Evidence: five dead checks in one session, several caught only by an independent reviewer mutating the deliverable rather than the base. Includes #550's silently-truncated runner (30 of 65 suites, reported green).

**2. Regex over prose may check exact positive anchors or structure. Never meaning, denial, or polarity — review the prose, or delete the marker (#493 req 6).**

[...]

Issue #556, opened 2026-08-10
Decide whether CI and the test suites need eval tests — or whether prose should not be tested at all


BaseInfinity:
## Scope card
- **One issue:** decide whether CI and the test suites need eval-style tests, and where.
- **Acceptance:** a written decision — which suites (if any) get evals, which explicitly do not and why, published on this issue. If the answer is "none", that ships as the answer.
- **Allowed paths:** research and this issue only. **No code, no new suites, no workflow edits** until the decision is approved.
- **Exclusions:** does not build an eval harness. Does not touch `.github/workflows/`, `hooks/`, or `scripts/`.
- **Risk tier:** low (research)
- **Estimated diff:** 0 lines of code

## Why now

The suites here are pass/fail bash assertions. They answer *"does this string exist"*. They do not answer *"does the agent behave correctly"* — which is the only thing that matters for a repo whose product is agent instructions.

Concrete evidence from the v1.97.0 session (2026-08-08/10):

- A guard on `skills/sdlc/SKILL.md` + `CLAUDE_CODE_SDLC_WIZARD.md` took **four review rounds** and was ultimately **deleted**. Every round found the same defect: a text-presence check cannot detect polarity. The final defeat was Markdown strikethrough — `~~the count never resets~~` leaves the searched string byte-identical while the live sentence teaches the opposite, and the suite stayed green.
- Five separate **dead checks** shipped or nearly shipped in one session: guards that passed on every input, including the inputs they existed to reject.
- #551 deleted a prose guard for the same reason. #493 requirement 6 already prefers deleting a mutable prose marker over guarding it.

The open question this issue exists to answer: **is the right response an eval, or is it "don't test this at all"?**

## The real question, stated fairly

The maintainer's position, which this issue must take seriously rather than route around:

> You don't need to write tests to protect documentation UNLESS the documentation may impact the agent (CLAUDE.md, AGENTS.md, skills). And even then some things aren't worth testing when they're hard to test — like meta changes to the harness.

So the decision is three-way, not two-way:
1. **Eval it** — the doc changes agent behavior and the behavior is observable. Worth an eval.
2. **Assert it** — the doc has a mechanical, structural property (a file exists, two copies are byte-identical, a JSON key is present). Plain assertions already work here.
3. **Neither** — the doc is prose whose correctness is a judgement call. Review it; do not guard it.

Category 3 is the one currently missing, and it is where four rounds went.

## What to investigate

[...]

Issue #542, opened 2026-08-10
Pilot: measure whether the rules work, with a pre-registered failure signal


BaseInfinity:
## Scope card
- **Acceptance:** 10 eligible reviews instrumented; result published to #534 and codex-sdlc-wizard#110 whichever way it falls.
- **Allowed paths:** measurement notes only. **No code.**
- **Exclusions:** do not build tooling to collect this. `git log`, `git diff --numstat`, `git blame` and existing artifacts are sufficient.
- **Risk tier:** low
- **Estimated diff:** 0 lines of code

## Why
Per EXISTENCE RULE, the rules shipped so far have not been observed producing either outcome. This is the observation.

**Pre-registered failure signal (agreed with both reviewers, recorded before the data):**
> The same root task reaches reviewer invocation 3, or gains a new mechanism/promise, without a continue/stop decision and cost line recorded BEFORE that invocation. Second signal: a blocker blamed to a PR-created line is repaired with net-new machinery instead of the accretion being deleted.

Sol's sharpening, adopted verbatim: **if nothing reaches pass 2 during the window, the result is inconclusive — not evidence that prose worked.**

**Also record per dual review:** each reviewer's confidence BEFORE and AFTER reconciliation. The Codex host measured Sol 86->89 and Fable 70->88; the reviewer that moved most started least confident. Pooling both repos' numbers answers *when is a reconcile round worth paying for* empirically. A reconcile round where neither number moves is one to stop paying for.

## Blocks
The review-budget runner does not get built until this pilot reports.

BaseInfinity:
## Instrumentation rewritten — both failures stalled BEFORE the gate, not at it

The completion-boundary question is answered with measured data, and it changes this pilot's design.

**The gate is not the bottleneck.** VisualHUD's five closed issues crossed the completion boundary in **~5, 5, 6, 6 and 26 minutes** from final commit to closure. Both failures stalled *before* a stable candidate ever existed:
- **VisualHUD #15** — never froze: review kept generating findings in unrequested machinery it had itself introduced.
- **m180 `dff6867`** — froze, but on an overbroad three-issue unit whose own acceptance doc reported **1 Ready cell out of 48, 22 Blocked**, and explicitly stated #17/#19 remained open. The reviewers were being used as implementation-discovery machinery on a known-blocked candidate.

[...]

PR #560, opened 2026-08-10
feat(sdlc): cumulative pass budget per root task, and a scope card in PLANNING (#539, #538)

### LINKED ISSUE #538: Scope card: the artifact that makes CLOSED ALLOWLIST checkable
## Scope card
- **Acceptance:** every issue template and the `/sdlc` planning phase produce a scope card BEFORE work starts: one issue, acceptance criteria, allowed paths, exclusions, risk tier, estimated diff.
- **Allowed paths:** `.github/ISSUE_TEMPLATE/`, `skills/sdlc/SKILL.md`, `cowork/skills/sdlc/SKILL.md`, `CLAUDE_CODE_SDLC_WIZARD.md`, `tests/`
- **Exclusions:** no hooks, no runtime enforcement, no new parsers. Documentation + tests only.
- **Risk tier:** low (docs + guards)
- **Estimated diff:** ~80 lines deliverable, ~120 lines test

## Why
Adopted from BaseInfinity/codex-sdlc-wizard#110. Our CLOSED ALLOWLIST asserts "the issue's requested behaviors are the whole job" but never named the artifact that makes it checkable. Without the card there is nothing to compare growth against, which is why nothing fired during #520.

The card makes the breaker mechanical — it trips on **a new subsystem/path/criterion, >2x the estimate, or two corrective rounds**.

## Test
A guard asserting the skill's planning phase requires a scope card, plus a negative fixture proving the guard stays silent when one is present. Per EXISTENCE RULE, both outcomes observed.

### LINKED ISSUE #539: Pass budget must be cumulative per root task, not per frozen scope
## Scope card
- **Acceptance:** the handoff/review artifact records `root_task`, `base_sha`, and a pass count that NEVER resets when scope is re-frozen. New scopes append child records.
- **Allowed paths:** `skills/sdlc/SKILL.md`, `cowork/`, `CLAUDE_CODE_SDLC_WIZARD.md`, `tests/`
- **Exclusions:** no gate/hook changes in this issue.
- **Risk tier:** medium (it is the termination rule)
- **Estimated diff:** ~40 lines deliverable

## Why
Sol's Q4 finding, confirmed by Fable: **"frozen scope" is redefinable.** An agent can re-freeze after every fix, and "two passes per frozen scope" silently becomes unlimited. Root = verbatim request + base SHA.

This also guards the incremental-commit model (codex-sdlc-wizard#109): if each commit's review resets the count, incremental review becomes 20 rounds under a new name.

## Test
A fixture that re-freezes scope 3x and asserts the cumulative count still reads 3, not 1.

BaseInfinity:
Closes #539. Closes #538.

## #539 — the pass budget is cumulative per root task

The budget was countable per *re-frozen scope*, so re-freezing handed the driver
a fresh budget indefinitely. It is now counted per **root task** — the verbatim
request plus the base SHA — and the count never resets. New scopes append to the
record rather than replacing it.

## #538 — PLANNING requires a scope card

CLOSED ALLOWLIST said "the issue's requested behaviors are the whole job" but
never named the artifact you compare growth against. That is why nothing fired
during #520. PLANNING now requires a six-field card — one issue, acceptance
criteria, allowed paths, exclusions, risk tier, estimated diff — and all three
issue templates carry it.

## What the tests do and do not prove

They are **presence-only** and every message says so. They check that the field
names and clauses are present; they do not assert the surrounding prose teaches
the correct rule.

That is the outcome of four review rounds, not a shortcut. A definition
assertion pinning the rule's wording was written and defeated four consecutive
times, finally by Markdown strikethrough — `~~the count never resets~~` keeps the
pinned string byte-identical while the live sentence teaches the opposite, and
the suite stayed green. A text-presence check can detect presence, never
polarity. A pre-registered tripwire fired and the assertion was retired.

Seven blocking findings across four rounds; **all seven were in the test, zero in
the 23-line deliverable**, which was correct at round 1 and never changed.
Residual risk is filed as #556 and #557.

## Verification

- `tests/test-doc-consistency.sh` — 137/137
- `tests/test-cowork-drift.sh` — 31/31
- `skills/sdlc/SKILL.md` and `cowork/skills/sdlc/SKILL.md` byte-identical, SHA-256 `81093a6e…d74aa`
- `bash -n`, `git diff --check` clean
- Seven changed files, all within the scope card's allowed paths

Both reviewers ran mutation batteries including the strikethrough reversal and
an append-polarity reversal, and confirmed the surviving messages are literally
true of a document teaching the opposite rule.


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

**Reviewer:** Codex CLI — GPT-5.6 Sol, `model_reasoning_effort=high`
**Rounds:** 6, full dialogue with per-finding responses recorded in `.reviews/handoff.json`
**Verdict:** CERTIFIED — confidence 99/100

Round 6 certify condition and result:

- Swept all 22 result messages in the added block; each is presence-only and literally matches its check.
- Compared against the exact round-5 transcript snapshot: zero non-message changes. Only the four specified result strings changed; no check, regex, or pattern changed.
- `test-doc-consistency.sh`: 137/137 passed.
- `test-cowork-drift.sh`: 31/31 passed.
- Skill copies byte-identical, SHA-256 `81093a6e...d74aa`.
- `bash -n` and `git diff --check` passed.
- Exactly seven changed files, all allowlisted.

Round history: R1 four dead checks; R2 whole-file wizard reversal stayed green; R3 whole-file skill reversal stayed green; R4 Markdown strikethrough defeated the exact pins with every pinned string byte-verbatim, firing the pre-registered tripwire — retire the definition assertion, do not extend the patterns; R5 four retained messages overstated what their grep proves; R6 clean.

Pilot confidence record for #542: R4 82 to 98, R5 96 to 99, R6 92 to 99.

```json
{"reviewer": "codex-gpt-5.6-sol-high", "verdict": "YES", "confidence": 99, "sha": "3d8163fdfdc4e46ebe5a340d965a62d4a82151a7"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

**Reviewer:** Fable, `xhigh`, fresh subagent scoped to the diff
**Rounds:** 6, independent of Sol, with a mutation battery each round
**Verdict:** CERTIFIED — confidence 97/100

Round 6 certify condition and result:

- All result messages swept, not only the four rewritten. Re-ran the round-5 mutation that produced the P1: the wizard clause changed to append child records by replacing prior records, a document teaching REPLACE. 137/137 green, and the message that now prints is literally true of that document, since it claims only that the words co-occur in order and explicitly disclaims polarity. Round-5 P1 closed. Wizard restored, four SHAs verified.
- No check, regex, or pattern changed. Verified against a round-4 snapshot rather than taken on faith: 95 diff lines, consisting solely of the already-cleared retirement plus the four message strings. The only grep-bearing line in the whole diff sits inside the deleted helper. Line-number triangulation confirms the message edits were line-count preserving.
- 137/137 doc-consistency, 31/31 cowork-drift, byte parity at SHA-256 `81093a6e...d74aa`, `bash -n` and `git diff --check` clean.
- Exactly the seven allowlisted files.

Two P3 no-action observations recorded as accepted residual alongside #557: the breaker test's messages attribute a whole-file grep result to the breaker without a polarity disclaimer, and one absence-direction message is near-entailed but order sensitive. Both strings are unchanged since round 4 and were swept by both reviewers under this exact criterion without a flag; re-reading unchanged text more strictly is the ratcheting the tripwire exists to stop.

The 3 points off 100: the round-5 baseline is reconstructed rather than byte-exact, plus the two P3s above.

Pilot confidence record for #542: R5 80 to 96, R6 85 to 97, including a downward self-recalibration after a miss.

```json
{"reviewer": "fable-xhigh", "verdict": "YES", "confidence": 97, "sha": "3d8163fdfdc4e46ebe5a340d965a62d4a82151a7"}
```

PR #562, opened 2026-08-10
docs(sdlc): fixed means observed — name the observable before reporting a fix

### LINKED ISSUE #525: SDLC gap: config/env changes bypass the cross-model gate, and "fixed" is allowed to mean "edited"

BaseInfinity:
Closes #525.

A settings fix was reported fixed on 2026-08-08 while the bug was still live. The evidence for "fixed" was that the edit had been made — nothing had been observed. The answer was already sitting in `.reviews/` from PR #468.

Adds one section: name the observable that would differ if the change were NOT fixed, then go look at it. Settings file edited → read the live process env, not the file. Hook changed → fire it. Threshold changed → measure it.

Plus the corollary that makes this class recur: **out-of-repo changes get no gate.** Global `settings.json`, env vars, shell rc, scheduler entries produce no diff, no PR, and no reviewer. They need the verification command stated in the same message as the change, and if a live process won't pick up the edit, that has to be said instead of "fixed".

18 additive lines across `skills/sdlc/SKILL.md`, the byte-identical `cowork/` copy, and the wizard mirror.

**Tests: none, deliberately.** Bucket three of the three-way call — prose whose correctness is a judgement call. Cross-model review is the guard, and both legs ran. `test-doc-consistency.sh` 137/137 and `test-cowork-drift.sh` 31/31 already cover the mechanical contract (byte parity).

**Both reviewers certified at round 1 with zero findings at every severity** — the first zero-finding review of this cycle. Sol independently verified that PR #468's review artifact predates the incident and confirmed `Math.floor(window * (PCT_OVERRIDE / 100))` against Claude Code v2.1.221. Fable independently reproduced both verifications from primary sources and swept all shipped prose for contradicting rules, finding none.

Merging on `--user-approved` rather than `--dual-certified`: the gate requires `round >= 2`, and there were zero findings for a round 2 to be about. Filing that as its own defect.

### DIFF
--- a/CLAUDE_CODE_SDLC_WIZARD.md
+++ b/CLAUDE_CODE_SDLC_WIZARD.md
@@ -2781,6 +2781,12 @@ Reproduce → Isolate → Root Cause → Fix → Regression Test
 
 **Environment-specific bugs** (works locally, fails in CI/staging/prod): Check environment differences (env vars, OS version, dependency versions, file permissions). Reproduce the environment locally if possible. Add logging at the failure point — don't guess, observe.
 
+### Fixed Means Observed
+
+Before reporting a fix, name the observable that would differ if it were NOT fixed — then go look at it. Settings file edited → read the live process env, not the file. Hook changed → fire it. Threshold changed → measure it against the real files. If the only evidence is "I made the edit", the state is *submitted*, not fixed.
+
+**Out-of-repo changes get no gate** (global `settings.json`, env vars, shell rc, scheduler entries): no diff, no PR, no reviewer ever sees them. Before editing, check `.reviews/` artifacts and memory for prior findings on the subject; state the verification command in the same message as the change; if a live process won't pick up the edit (env vars need a restart), say so instead of "fixed". **Evidence:** 2026-08-08 (#525) — a settings fix was reported fixed while the bug was still live; the answer was already in `.reviews/` from PR #468.
+
 ## Step 8: Create CLAUDE.md
 
 Create `CLAUDE.md` in your project root. This is your project-specific configuration:
--- a/cowork/skills/sdlc/SKILL.md
+++ b/cowork/skills/sdlc/SKILL.md
@@ -189,6 +189,12 @@ Mandatory steps:
 
 Reproduce → Isolate → Root Cause → Fix → Regression Test. No skipping. `git bisect` for regressions. 2 failed attempts → escalate (Confidence Check), not straight to the user.
 
+## Fixed Means Observed
+
+Before reporting a fix, name the observable that would differ if it were NOT fixed — then go look at it. Settings file edited → read the live process env, not the file. Hook changed → fire it. Threshold changed → measure it against the real files. If the only evidence is "I made the edit", the state is *submitted*, not fixed.
+
+**Out-of-repo changes get no gate** (global `settings.json`, env vars, shell rc, scheduler entries): no diff, no PR, no reviewer ever sees them. Before editing, check `.reviews/` artifacts and memory for prior findings on the subject; state the verification command in the same message as the change; if a live process won't pick up the edit (env vars need a restart), say so instead of "fixed". **Evidence:** 2026-08-08 (#525) — a settings fix was reported fixed while the bug was still live; the answer was already in `.reviews/` from PR #468.
+
 ## Release Planning (Task Ships a Release)
 
 All ROADMAP items planned at 95%, dependencies identified, presented together (catches conflicts), pre-release CI audit across merged PRs — a green checkmark is insufficient. User approves, then implement in priority order.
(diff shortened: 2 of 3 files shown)


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

SHAPE: SOUND. Zero findings at P0, P1, P2, and P3.

Verified the evidence cite rather than accepting it: PR #468's review artifact was created July 24, before the August 8 incident, and contained the relevant autocompact answer. Claude Code v2.1.221 confirms the multiplication semantics directly. No shipped surface contradicts the new section — Rule 3, deploy verification, and the autocompact guidance are complementary, and the Post-Mortem test formula is correctly deferred to the newer three-way rule. Skill and cowork copies byte-identical; wizard mirror verbatim at the intended heading level. 137/137 and 31/31. Exactly 18 additions across the three allowlisted files, no deletions, no tests, hooks, or guards touched.

```json
{"confidence": 99, "reviewer": "codex-gpt-5.6-sol-high", "sha": "ce44f582475ad3f0bdf07b79234b38a1bef25cc9", "verdict": "YES"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

SHAPE: SOUND. No findings at any severity.

Checked the alternative homes fresh rather than trusting the spec: Debugging Workflow is test-centric, Post-Deploy Verification is deploy-scoped template content, Confidence Check is pre-approach — none fits a fix with no test and no deploy. The section exists in the right form. Independently reproduced both of the other leg's verifications from primary sources in this repo: artifact provenance predating the incident, and the multiplication semantics. Swept every shipped surface for contradicting fix-reporting rules, including AI_SETUP_LANES.md — zero found. Byte parity confirmed by cmp; suites green; working tree carries exactly the three allowlisted files.

Residual risk is prose judgement that no reviewer closes by inspection — bucket three of the three-way call, where cross-model review is itself the guard. Both legs have now run.

```json
{"confidence": 97, "reviewer": "fable-xhigh", "sha": "ce44f582475ad3f0bdf07b79234b38a1bef25cc9", "verdict": "YES"}
```

BaseInfinity:

**Reason:** Standing blanket merge authorization. Two independent round-1 certifications with zero findings at every severity (Sol 99, Fable 97), both posted as parseable clearances. Gate defect for the round>=2 case filed as #563.

- ` clearance-artifact checks (.reviews/merge-clearance-562.json)`

Head: `ce44f582475ad3f0bdf07b79234b38a1bef25cc9`

Issue #569, opened 2026-08-10
Version-marker guards are hand-enumerated: four stale markers passed every suite during the v1.97.0 prep


BaseInfinity:
During the v1.97.0 release prep, four version markers were left at 1.96.0 and **every suite passed**:

- `cowork/README.md:131` — "Tracks the main wizard version: **1.96.0**"
- `ROADMAP.md:18` — "Last release: v1.96.0"
- `ROADMAP.md:20` — "Prior release: v1.95.0"
- `ROADMAP.md:207` — the living tracker row, whose own text says "bump every release"

The v1.96.0 release commit rolled all four forward, so they are maintained surfaces, not dead text. Review caught them; no guard did.

**This is #493's exact class, for the third time.** That issue records three markers sitting at v1.90.0 across three releases and three attempts at a guard, each proven vacuous. The guards that exist cover `package.json`, the plugin manifests, `SDLC.md`, and the wizard doc's HTML comments — the set someone once enumerated by hand. Every marker outside that hand-list is unguarded, and the failure is silent.

Same shape as #561's registration surfaces, and the same fix applies: **stop enumerating the places a version can appear, and quantify over a set the repo owns.** Candidate: sweep every tracked shipped file for a version-shaped string and require each occurrence to be either current, or explicitly annotated as a historical reference.

The annotation half is load-bearing. `ROADMAP.md` legitimately contains "v1.96.0" as a *prior release* record and in a triage note, and both must stay. A guard that cannot distinguish "stale marker" from "historical fact" will either fail on correct text or be narrowed until it catches nothing — which is how the previous three attempts died.

Note the fix must land with an observed RED against the *nearest wrong form*: a marker left stale, not a marker deleted. Red-by-absence is what let the earlier attempts ship vacuous.

Related: #493 (the original), #561 (the enumeration-vs-globbing method ruling).

BaseInfinity:
Closed. Reviewer's words: *"#569 proposes another semantic-ish version guard after this repo repeatedly proved that pattern costly."*

The counterpart yielded on exactly this point, citing this repo's own history: **#493 was three guard attempts, all vacuous.** A hand-enumerated version-marker guard is the same pattern that has already failed here more than once, and the enumeration problem it complains about is structural — you cannot enumerate an externally-owned set.

Stale version markers get caught by release verification, where a human reads the release. Not worth an eighth dead check.

Closed on reconciled cross-model triage, 2026-08-10.

Issue #530, opened 2026-08-09
/sdlc must carry the standing setup itself — the maintainer is retyping the same instructions every session


BaseInfinity:
## The complaint, verbatim

> "if u cant keep using fable as ur brain then need to improve /sdlc cuz im tired of typing the same fucking commands"

and, on the same session:

> "why do you keep stopping huh? why do you keep HITL me when I dont need to huh?"
> "why the fuck do you keep interrupting this is worse than /goal mode"
> "i am getting 0 done with this repo and im getting so much more done with the same harness in codex"

## The mechanism

`advisor()` — the one-call path to Fable — is **server-side disabled repo-wide** by an Anthropic rollout, not broken locally. The documented fallback is spawning a Fable subagent. That works, but it is not equivalent:

| | `advisor()` | Fable subagent |
|---|---|---|
| Context | Entire transcript, automatic | Must be hand-assembled into the prompt every time |
| Cost to invoke | One no-arg call | A several-hundred-word brief |
| Failure mode | — | Agent gets a summary, not the evidence; rules on a strawman |

So every design consult costs a hand-written brief. The maintainer notices this as "typing the same commands," and the agent notices it as friction that biases toward *not* consulting — which is exactly the failure `#526` (escalate to design after the second repair) exists to prevent. **A fallback that is 20x more expensive to invoke than the primary is a fallback that quietly stops being used.**

## Two distinct gaps

**1. The standing setup is not in the skill.** Fable-as-brain / Codex-as-adversarial-check, effort per model, "never ask what a model can settle," the escalation ladder — these live in `CLAUDE.md`, in the maintainer's memory files, and in repeated verbal instruction. Some are in `skills/sdlc/SKILL.md`; the ones that keep getting restated are not. Anything the maintainer has to say twice is a skill defect, not a discipline problem.

**2. The subagent brief is unautomated.** There is no `advisor-fallback` helper. Every consult is bespoke prose. Candidate: a documented brief template (evidence / the contradiction / numbered design questions / "read the files, do not trust my summary") — the shape that produced a decisive ruling on #520 — so the cost of consulting drops to filling in slots.

## Why this is a MECHANISM issue, not a discipline one

This is the same finding as `#519` ("Autonomy is a MECHANISM gap, not a discipline gap: an instruction cannot start a new turn"). Adding another sentence to the skill saying "consult Fable early" does nothing, for the same reason: an instruction cannot lower the cost of an action. Only tooling can.

[...]

PR #570, opened 2026-08-10
release: v1.97.0 — the hook that blocked the maintainer is gone, and TDD RED is scoped

BaseInfinity:
Ships v1.97.0. Seven issues are merged and undelivered — **merging delivered none of them.** Cowork plugins run from a local cache, and the maintainer's is still v1.93.0 with the `UserPromptSubmit` classifier live. The tag is what moves it.

## What a consumer gets

**The hook that denied its own maintainer, twice, is deleted.** The second denial came *after* a repair that a passing test had pinned — which is the case for deleting rather than narrowing a third time. The enforcement cost is stated in `cowork/README.md`, not hidden: in Cowork, a skip-the-process prompt followed by edits to existing files now hits no gate at all.

**TDD RED is scoped to where a RED mutation is writable**, with the three-way call — EVAL it, plain-assert it, or don't test it and let cross-model review be the guard. Six guards had shipped that passed on every input including the ones they existed to reject.

**A guard that covers surfaces nobody knew existed.** Hook-registration absence is now checked over the shipped file set rather than a hand-list of mechanisms, failing closed on unparseable formats. Review then found that Anthropic also documents hooks in agent and command frontmatter — a fifth and sixth surface no enumeration across six rounds ever named — and the walk already covered both.

Plus: "fixed" means observed (#525), the cumulative pass budget and planning scope card (#539, #538), the autocompact guidance fix (#531), `/sdlc` carrying the standing setup (#530), the deleted 20,000-byte ceiling gate (#489), and native-installer guidance (#476).

## Review found two P1s, both mine

**Version parity was incomplete after I declared the sweep complete.** Four markers still read 1.96.0 — `cowork/README.md` and three in `ROADMAP.md`, all maintained surfaces that the v1.96.0 release commit had rolled forward — and **every suite passed anyway.** That is #493's class for the third time: the guards cover a set someone once enumerated by hand, and everything outside it fails silently. Filed as **#569**, with the enumeration-vs-globbing treatment #561 just proved out. Not fixed here; it is not release work.

**The CHANGELOG was wrong in four places.** I compressed #531, #530 and #489 into a single line and misdescribed all three — #489 *deleted* the unmeasured ceiling gate rather than adding one — omitted #476 entirely despite it changing the shipped `README.md`, and claimed seven historical dead guards where six is the evidenced number. Now four accurate entries, and the seventh guard is stated as what it actually was: found *inside* #567, in the guard protecting the hook deletion, caught by a reviewer mutating the file rather than reading the assertion.

Both fixed and re-certified at 99%.

## Verification

`doc-consistency` 137/137, `release-drift` 21/21, `roadmap-integrity` 5/5, `cowork-drift` 30/30. `npm pack --dry-run` contains 25 expected files with every `files` entry present. `test-persist-score-history.sh` fails identically on `main` — pre-existing, confirmed by merge-base repro, tracked as #566.

An isolated working-tree install reports exactly `Hooks (1)  PreToolUse`; the real v1.93.0 cache reports two. That difference is the release.

## After merge

`git tag v1.97.0 && git push origin v1.97.0`. Four documents in this release hard-pin the literal string "v1.97.0" and ship true only once the tag exists.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -6,7 +6,7 @@
   },
   "metadata": {
     "description": "SDLC enforcement plugins for AI-assisted development",
-    "version": "1.96.0"
+    "version": "1.97.0"
   },
   "plugins": [
     {
@@ -16,7 +16,7 @@
         "repo": "BaseInfinity/claude-sdlc-harness"
       },
       "description": "SDLC enforcement for AI agents \u2014 TDD, planning, self-review, CI shepherd",
-      "version": "1.96.0",
+      "version": "1.97.0",
       "author": {
         "name": "Stefan Ayala"
       },
@@ -36,7 +36,7 @@
         "path": "cowork"
       },
       "description": "SDLC enforcement for Claude Cowork \u2014 TDD, planning, self-review via skills and one prompt-based hook",
-      "version": "1.96.0",
+      "version": "1.97.0",
       "author": {
         "name": "Stefan Ayala"
       },
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "sdlc-wizard",
-  "version": "1.96.0",
+  "version": "1.97.0",
   "description": "SDLC enforcement for AI agents \u2014 TDD, planning, self-review, CI shepherd",
   "author": {
     "name": "Stefan Ayala",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,92 @@ All notable changes to the SDLC Wizard.
 
 > **Note:** This changelog is for humans to read. Don't manually apply these changes - just run the wizard ("Check for SDLC wizard updates") and it handles everything automatically.
 
+## [1.97.0] - 2026-08-10
+
+### Removed — a hook that denied its own maintainer, twice
+
+- **The Cowork `UserPromptSubmit` prompt classifier is deleted.** It blocked the maintainer's
+  own instruction to codify a policy change, classifying a *description* of policy as an attempt
+  to *evade* it. That was the second denial, not the first: on 2026-08-07 it refused "do the
+  prose rename now, dont TDD something like this... just verify after" — a stated reason plus a
+  stated alternative, a safeguard being **substituted**, not removed. A justified-exception
+  carve-out shipped in response and a test pinned its text. The test passed. The hook denied the
+  maintainer anyway.
+
+  No prompt wording guarantees an LLM classification outcome, so it is unsatisfiable by
+  narrowing, and it is unfalsifiable from inside — reporting the false positive requires a
+  prompt it may block. Same class and same remedy as the `Stop` hook removed in v1.92.0.
+  Blocking hooks fire on **acts** (`PreToolUse`), never on turn-level subject matter. (#561)
+
+- **The honest cost, stated rather than hidden.** In Cowork, a prompt asking to skip planning or
+  review, followed by edits to files that already exist, now hits **no gate at all** —
+  `PreToolUse` fails open for edits, and the CLI's commit and merge gates cannot run there.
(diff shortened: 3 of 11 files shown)


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Certified. No findings at head `5016eb6`; cleared for tagging.

Round 1 returned two P1s, both in this release's own bookkeeping: version parity was incomplete after the sweep had been declared complete — four maintained markers still at 1.96.0, with every suite passing — and the CHANGELOG misdescribed three merged issues, omitted a fourth that had changed the shipped README, and claimed seven historical dead guards where six is evidenced. Round 2 certified the fixes at 99%.

This round covers the two sentences the second leg's P2 produced, which round 2 predates. Both corrected claims hold, verified against primary history rather than the fix's own framing: v1.84.0 is the first tagged Cowork plugin, every release through v1.96.0 contains `UserPromptSubmit`, v1.97.0 removes it, and the seventh guard caught populated re-adds while missing an empty one. The diff is surgical.

Earlier verifications stand: an isolated v1.97.0 working-tree install reports exactly `Hooks (1)  PreToolUse` while the real v1.93.0 cache reports two; `npm pack --dry-run` contains 25 expected files with every `files` entry present; all seven commits in `v1.96.0..c20616c` are represented; `test-persist-score-history.sh` reproduces identically on detached clean `main` (#566, pre-existing).

```json
{"confidence": 99, "reviewer": "codex-gpt-5.6-sol-high", "sha": "5016eb697f185ffdf44d4a76b5c1f9ffdbf3dc7e", "verdict": "YES"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Certified, pinned to `5016eb6`.

Found one P2 in a paragraph two review rounds had already read past: the delivery note promised relief once a cache refreshes "past v1.93.0". The classifier was added at `13c6a94` and removed only at `c20616c`, so it ships in every Cowork version through v1.96.0 — a consumer landing on v1.94, v1.95 or v1.96 is past v1.93.0 and still has the hook that denied its maintainer twice. v1.93.0 was the driver's own installed cache version, leaked into the one paragraph telling other people whether they are fixed. In a release whose two certification legs both existed to make this document true, that sentence was not certifiable.

The fix holds and its strengthened claim survives independent checking: the cowork plugin manifest and the classifier are first contained in the same tag, v1.84.0, so "every version" has no false members at the old end, and `git log -S` shows exactly two occurrence-changing commits — continuous presence, no gap, so no version inside the range is falsely accused either.

Verified independently rather than re-deriving leg 1: the verification command against the live v1.93.0 cache reports `Hooks (2) PreToolUse, UserPromptSubmit`, and post-update `Hooks (1) PreToolUse` — the exact 2 to 1 observable the entry promises. All four corrected entries match their commits. Version parity holds in both directions, with three remaining 1.96.0 references all correctly historical. `npm pack` at 1.97.0, 25 files matching the ships-list.

On the judgement call: the framing is honest and nothing should be held back. The entry leads with self-indictment — the deletion, then a bullet naming the no-gate-at-all consequence, then an admission that the mandate produced fake testing including a seventh dead guard inside this very release. Read as a consumer who left because this tool made them circle, every deliverable here reduces circling.

```json
{"confidence": 97, "reviewer": "fable-xhigh", "sha": "5016eb697f185ffdf44d4a76b5c1f9ffdbf3dc7e", "verdict": "YES"}
```

BaseInfinity:

**Reason:** Release v1.97.0. Standing blanket merge authorization, plus the maintainer explicitly requested this release tonight and is blocked on it for a multi-repo trial. Dual cross-model certification at head 5016eb6: Sol 99, Fable 97, both posted as parseable clearances. Version bump is the point of the PR, not incidental — nine files, verified parity in both directions, all seven merged issues represented in the CHANGELOG.

- ` package.json version bump (release-adjacent confirmation)`
- ` clearance-artifact checks (.reviews/merge-clearance-570.json)`

Head: `5016eb697f185ffdf44d4a76b5c1f9ffdbf3dc7e`

PR #576, opened 2026-08-11
fix(tests): make the score-history fixture fail closed instead of running git in the caller's repo (#566)

### LINKED ISSUE #566: test-persist-score-history.sh fails open: its mktemp/cd fallback can run git against the real repo

BaseInfinity:
Closes #566.

`tests/test-persist-score-history.sh` allocated its sandbox with a bare `root=$(mktemp -d)` and checked nothing. When mktemp fails — an unwritable TMPDIR under sandbox does it — `root` is empty and the fixture carried on in whatever directory it was invoked from, running `git init`, `git commit` and `git push` **there**. That directory was the real repo. Not hypothetical: it happened to the maintainer.

**RED, observed** (PATH-shadowed failing mktemp): the fixture created `ci-clone/`, `ci-clone/README.md` and `ci-clone/tests/e2e/score-history.jsonl` in the caller's working directory.
**GREEN:** caller's directory untouched, exit non-zero, fixture still 5/5.

## `cd ""` is version-dependent, and the default shell has the dangerous one

| shell | `cd ""` |
|---|---|
| **bash 3.2** (`/bin/bash` on macOS — what every `#!/bin/bash` script gets) | **returns 0, prints nothing, does not move** |
| bash 5.x | returns 1, `cd: null directory` |

So on macOS `cd "$root" || exit 1` does **not** fire on an empty root, which makes the per-call-site guard load-bearing there. On Linux CI the cd guard does fire. Both are kept.

That fact was nearly mis-retracted mid-review: a re-check under Homebrew bash 5 contradicted the original bash 3.2 observation and the true claim was briefly withdrawn as false. **"I tested it" is only evidence when you tested the interpreter the code actually runs under** — which was also the guard test's own defect: it invoked the fixture as `bash <file>`, picking up PATH bash 5.x and silently testing a shell nobody uses. It now executes the file so the shebang decides.

## Review

One round, two P1s, both fixed. The first was a **dead check** — an always-failing mktemp aborts at the *first* allocation, so sites 2–4 were never reached and their guards never exercised. Proved by deleting a guard and watching the mutant still pass. The stub now fails on a *selected* invocation so each site takes a turn.

**Known unknown, stated rather than papered over:** with any single call-site guard removed, the fixture still exits 1 and still pollutes nothing, so the guard cannot currently distinguish that mutation. `exit` inside `$()` was verified *not* to propagate to the parent, so the mechanism is not understood and is **not claimed**. The guard asserts the observable that matters — files in the caller's directory — and that is clean at all four allocation points.

## Scope

The **200+ other unguarded `mktemp -d` sites** across `tests/` are deliberately not here. Converting them in the PR that establishes the pattern is the fix-churn this cycle is cutting.

Guard wired into CI. fail-closed 9/9, persist-score-history 5/5.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,6 +225,9 @@ jobs:
       - name: Run persist score history tests
         run: ./tests/test-persist-score-history.sh
 
+      - name: Run fixture fail-closed tests (#566)
+        run: ./tests/test-fixtures-fail-closed.sh
+
       - name: Run local shepherd tests
         run: ./tests/test-local-shepherd.sh
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,6 +45,7 @@ Thank you for your interest in improving the SDLC Harness!
    ./tests/test-memory-audit-protocol.sh && \
    ./tests/test-community-paths.sh && \
    ./tests/test-persist-score-history.sh && \
+   ./tests/test-fixtures-fail-closed.sh && \
    ./tests/test-local-shepherd.sh && \
    ./tests/test-repo-complexity.sh && \
    ./tests/test-prompt-hook-fires-once.sh && \
--- a/tests/test-fixtures-fail-closed.sh
+++ b/tests/test-fixtures-fail-closed.sh
@@ -1,0 +1,156 @@
+#!/bin/bash
+#
+# GH #566 — test fixtures must FAIL CLOSED when their sandbox cannot be created.
+#
+# The bug this guards, stated as the failure it produced:
+#
+#   root=$(mktemp -d)     # fails under sandbox -> root is EMPTY
+#   mkdir -p "$root"      # mkdir -p "" -> fails, status ignored (no set -e)
+#   cd "$root"            # cd "" -> fails ("null directory"), status ignored
+#   git init ...          # runs in the CURRENT directory. The real repo.
+#
+# Both guarded lines FAIL and both statuses were discarded, so execution walked
+# straight past them into git commands aimed at the caller's working directory.
+#
+# `cd ""` IS VERSION-DEPENDENT, and the dangerous version is the default one:
+#
+#   bash 3.2  (/bin/bash on macOS - what every #!/bin/bash script gets)
+#             -> returns 0, prints nothing, does not move.  SILENT.
+#   bash 5.x  -> returns 1, prints "cd: null directory".
+#
+# So on macOS `cd "$root" || exit 1` DOES NOT FIRE on an empty root, which makes
+# the per-call-site `root=$(new_root) || exit 1` the load-bearing guard there.
+# On Linux CI the cd guard does fire. Both are needed.
+#
+# This was nearly mis-retracted mid-review: a check run under Homebrew bash 5
+# contradicted the original bash 3.2 observation, and the true claim was briefly
+# withdrawn as false. Recorded because the lesson is the point - "I tested it"
+# is only evidence if you tested the same interpreter the code actually runs.
+#
+# These tests assert the OBSERVABLE consequence — files appearing in the caller's
+# working directory — not the presence of any particular guard string. A test
+# that greps the fixture for `|| exit 1` would pass on a fixture that guards the
+# wrong variable.
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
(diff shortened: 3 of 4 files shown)


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Fable leg, rounds 1–2, at `0d50c3a`.

Round 1 raised one **P1**: the new header of `tests/test-persist-score-history.sh` asserted `cd ""` fails and returns 1, contradicting the header of `tests/test-fixtures-fail-closed.sh` in the same PR, which states that bash 3.2 — `/bin/bash` on macOS, and therefore the shebang shell — returns 0 silently and does not move, while only bash 5.x returns 1. The fixtures header is correct; the incident ran under bash 3.2, so `cd "$root" || exit 1` would not have fired and `root=$(new_root) || exit 1` is the load-bearing guard on macOS.

Round 2 confirms the P1 fixed: the persist header now defers to the fixtures header and states the version split. No contradiction remains, no new findings. The guard suite re-confirms the claim empirically on this machine (9/9, including the `cd ""` assertion); `test-persist-score-history.sh` 5/5.

```json
{"confidence": 96, "reviewer": "fable", "sha": "0d50c3ada31fd1ef108ecad2f92b71a3eb4f1df1", "verdict": "YES"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Sol (gpt-5.6) `high`, rounds 1–2, at `0d50c3a`.

Pre-commit round on the original branch raised two P1s against the guard itself — an always-failing `mktemp` stub aborted the fixture at its first allocation so call sites 2–4 were never exercised, and running the fixture as `bash <file>` picked up PATH bash rather than the shebang shell. Both fixed: the stub now fails on a selected invocation with one run per allocation site, and the fixture is executed directly.

Rounds 1 and 2 on the final diff: CERTIFIED, no findings. Round 2 was a targeted recheck of the persist-header comment fix (the P1 Fable raised) plus the one-line `CONTRIBUTING.md` addition that turned CI green — verified fixed, surface not expanded.

```json
{"confidence": 97, "reviewer": "sol", "sha": "0d50c3ada31fd1ef108ecad2f92b71a3eb4f1df1", "verdict": "YES"}
```

BaseInfinity:
**DUAL CROSS-MODEL CERTIFIED MERGE**

**Cleared by:** fable (posted by @BaseInfinity), sol (posted by @BaseInfinity) — both bound to `0d50c3ada31fd1ef108ecad2f92b71a3eb4f1df1`

**Merge-evidence path(s) this authorised:**
- `.github/workflows/ci.yml (merge-evidence path)`

**Verified:** CI `validate` green across every run of that name; no net-removed test files; no `package.json` version bump; clearance artifact CERTIFIED at round >= 2 bound to this SHA; the executing merge script and its redirect hook byte-match `origin/main`.

**ATTESTED, NOT AUTHENTICATED.** Both clearances were posted by the same `gh` token, so this records that two distinct reviewers returned YES at >=95 — not that two independent principals did. A new workflow file can still mint a green required check; the compensating layer is that both reviewers read this diff.

_Posted by `scripts/merge-pr.sh --dual-certified` before merging._

Issue #574, opened 2026-08-11
Cut the post-review confidence percentages — they carry no signal


BaseInfinity:
Both review legs independently proposed this cut, and the freeze that was holding it has been lifted by both.

**The percentages do not discriminate certified-correct from certified-wrong.** From PR #572:

- Leg 1 certified at **100%**. Leg 2 then found a P1 it had passed over three times.
- Leg 2 certified at **96%**. Leg 1 then found a P1 that only surfaced by *running the command*.

Two certifications, two confidences, two defects that landed anyway. The number predicted nothing in either direction.

**What actually discriminated was whether the reviewer executed the thing.** That is #573's rule, and it replaces this ritual rather than supplementing it.

Cutting it also removes a per-round token cost and a per-round argument — reviewers currently spend output justifying a figure that carries no information.

## Scope

Remove the confidence percentage from the review protocol in `skills/sdlc/SKILL.md`, the wizard doc, and `cowork/` (byte-parity). **Keep the verdict** — CERTIFIED / NOT CERTIFIED is the decision-bearing field and stays. The merge gate's clearance payload keeps its numeric field for now; changing that schema is gate work, not prose work.

Ships in **PR-A**, the prose batch ruled on #558. Filed separately because an unfrozen cut with no issue number is how work gets lost.

BaseInfinity:
Shipped in PR #577 (e810910), the PR-A prose batch. Both legs certified after one structured round with a verbatim cross-feed; architecture verdict SOUND from both.

Issue #573, opened 2026-08-11
An instruction is not a claim: executed-or-deleted for documented commands


BaseInfinity:
## The rule

**A doc may not instruct the reader to run a command unless that exact command string was executed and its observed output recorded.** No labeling escape hatch. Labeling is for *claims* a reader evaluates; an instruction is *executed*, so the only honest label for an unexecuted instruction is deletion.

Two amendments from the planner's ruling:

1. **Placeholders.** "Exact string" means exact modulo declared placeholders, and every placeholder must be instantiated at least once in the verification run.
2. **The output must observe the promised behavior, not merely exit 0.** This is the load-bearing part. `✔ already at the latest version (1.97.0)` proves an identifier *resolves*; it does not prove the command *updates*. If you cannot observe the promised behavior, **narrow the instruction's claim to what you did observe** rather than shipping the wider promise.

A command that genuinely cannot be run (destructive, environment unavailable) may be **described** but not **instructed**.

## Why this needed to exist

PR #572 was 24 lines of documentation. It produced **five P1s, and every one was the same defect** — a claim about a command's behavior wider than any observation of it.

1. "There is no CLI update path." False — produced by `claude plugin --help | head -20`; the subcommand list is alphabetical and `install`/`list`/`uninstall`/`update` all sat below the cut.
2. Documented a CLI path inside a Cowork Desktop document. The two surfaces keep separate plugin state.
3. Reported a reviewer's Desktop observation as unreproduced. The driver searched the filesystem; the reviewer opened the application.
4. Told the reader to check the `Version:` line of `claude plugin details`. That command prints no `Version:` field — `grep -c Version` returns `0`. The label belongs to `claude plugin list`.
5. Shipped an update command **that does not work**:

```
$ claude plugin update sdlc-wizard-cowork
✘ Failed to update plugin "sdlc-wizard-cowork": Plugin "sdlc-wizard-cowork" not found
```

**Defect 5 is why "attach the evidence or label it" was not enough.** That doc *had* labeled its evidence honestly — it stated the semantics came from the CLI's help text and that the command had not been run — and the broken instruction shipped anyway. The existing rule was followed and did not bind, because it treats an instruction like a claim.

## The certification precondition

**A review leg may not certify a doc containing imperative commands without pasted execution output.** This is the enforcement half, and it is checkable by the *other* leg — discipline with teeth rather than prose.

[...]

PR #572, opened 2026-08-11
docs: scope the plugin update path by surface, and name the command that lies (#571)

BaseInfinity:
Closes #571.

Both docs covered install and neither covered update. A consumer whose plugin was stale had nowhere to look — and the command whose name reads like it updates the plugin does not.

## What ships

Two files, +30 lines, docs only. `cowork/README.md` gains an **Updating an installed plugin** section; `CLAUDE_CODE_SDLC_WIZARD.md` gains the same guidance at its plugin-recovery point.

**Every instruction is written to what was actually executed.**

| Step | What was observed |
|---|---|
| `/reload-plugins` | ✅ **moved a version** — 1.93.0 → 1.97.0, `Hooks (2)` → `Hooks (1)` |
| `claude plugin update <plugin>@<marketplace>` | resolves; run against an already-current install, so **not observed moving a version** |
| `claude plugin update <plugin>` (bare) | ❌ fails: `Plugin "sdlc-wizard-cowork" not found`. `--scope user` does not help |
| `claude plugin marketplace update` | ✔ success banner, refreshes the **catalog** only |
| `claude plugin disable` / `enable` | ✔ success banner, no re-resolve |

All on Claude Code 2.1.221. Verification reads the version off the **header line** — `claude plugin details` prints no `Version:` field at all; that label belongs to `claude plugin list` — plus the full component set as a whitelist, never the absence of one hook name (#561).

**Scoped by surface, because the CLI and Cowork Desktop keep separate plugin state.** On one machine, at one moment: Desktop reports `sdlc-wizard-cowork@claude-sdlc-wizard` at **1.90.0**; `claude plugin details` reports `@sdlc-wizard-marketplace` at **1.97.0**. So the CLI commands are labeled CLI-only, and the Desktop path points at its own plugin UI — marked **not verified end-to-end** rather than guessed at.

## Review: five rounds, five P1s, one defect

Every one was a claim wider than the observation behind it.

1. Asserted no CLI update path existed. From `claude plugin --help | head -20` — the subcommand list is alphabetical, so `install`, `list`, `uninstall`, and `update` all fell below the cut.
2. Documented a CLI path inside a Cowork Desktop document.
3. Reported the reviewer's Desktop observation as unreproduced. I searched the filesystem; the reviewer opened the application.
4. Told the reader to check the `Version:` line of `claude plugin details`. `grep -c Version` returns `0`.
5. **Shipped an update command that does not work** — the bare identifier fails.

[...]

Issue #564, opened 2026-08-10
Reviewers reconcile with EACH OTHER, verbatim — the driver relays, it does not merge


BaseInfinity:
## The gap

`skills/sdlc/SKILL.md` currently says, in full:

> **Multi-reviewer:** respond to each independently.

That is the *driver* answering two reviewers in isolation. The reviewers never see each other. It is the opposite of what actually works, and it is where rounds multiply: when two reviewers disagree, the driver merges their prose into a third position neither endorsed, which then needs its own round to settle.

The maintainer's correction, verbatim:

> when i say reconcile i want the models to reconcile with themselves (each other) then come back to you if i ask, i dont trust you to reconcile their words together but i trust them to and then i trust you to reconcile that

## The protocol that worked

Four phases. Run on #558, #559, and #561 in one cycle; every one converged.

**1. Independent.** Each reviewer reviews the artifact blind to the other. No shared draft, no summary of the other's position. Divergence here is the signal — two reviewers who saw each other's work first produce correlated findings and the second reviewer is worth nothing.

**2. Cross-feed, verbatim.** Each reviewer receives the other's position *as written*, not summarized. The driver relays; it does not synthesize. Paraphrase is where positions get subtly restated into something the author would not sign.

**3. Argue it out.** Each concedes what is right and holds what is wrong with repo-verifiable evidence — "a concession to a misread is worse than a hold." A hold needs a file, a line, a command output. Not an opinion.

**4. Converge, then present one thing.** They hand back a single reconciled position. The driver reports it. The human sees the settled answer and the surviving splits, not two transcripts to referee.

## Why the driver must not merge

The driver is the lowest-ranked participant in a review. It has the most context and the least independence — it wrote the thing. When it merges two reviewers' words it is grading its own work through a paraphrase it controls. The reviewers outrank it precisely because they did not write the code.

## Deadlock

A split that survives phase 3 routes by **originating surface**: a position that first appeared as a design question or a recorded scope decision belongs to the planner; one that first appeared in the verdict line, the defect list, or a recheck response belongs to the reviewer, and the lower verdict applies. When a split spans both, the verdict's surface governs. Only a split that survives *that* goes to the human — and it goes as one question, not two positions.

## Evidence this is not theory

[...]

PR #577, opened 2026-08-11
feat(sdlc): ship the review contract as one prose batch (#558, #557, #573, #574, #564, #556)

### LINKED ISSUE #558: Define the review contract: severity scale, what blocks, an architecture verdict, and feature creep is always an issue
## Scope card
- **One issue:** define the review contract — a severity scale, what blocks, an architecture verdict, and a no-feature-creep rule.
- **Acceptance:** all four present in `skills/sdlc/SKILL.md`, mirrored in `CLAUDE_CODE_SDLC_WIZARD.md`, cowork copy byte-identical. Both reviewers get the identical contract.
- **Allowed paths:** `skills/sdlc/SKILL.md`, `cowork/skills/sdlc/SKILL.md`, `CLAUDE_CODE_SDLC_WIZARD.md`
- **Exclusions:** **no regex guard greping for these sentences** (see #557 — that mechanism is retired for prose). No new counter. No tooling.
- **Risk tier:** medium — changes what blocks a merge across every consumer repo.
- **Estimated diff:** ~20 lines across three files

## The gap

The severity scale is **used** in four places (`CLAUDE_CODE_SDLC_WIZARD.md` :4046, :4111, :4141, :3807) and **defined in none**. The wizard already records the resulting incident at :4220 — 2026-07-29, one reviewer was given P0–P3 and the other HIGH/MEDIUM/LOW on the same change, and two genuine reviews looked like a disagreement.

## 1. The scale

**Severity is impact if shipped. It is never how hard the fix is, and never which round the finding appeared in.** Without that clause the scale gets gamed in both directions: late findings inflated to justify another round, or deflated to close one out.

| | |
|---|---|
| **P0** | Stop the world — prod broken, data loss, secret leaked. Preempts the current task. |
| **P1** | This PR does not merge. It doesn't work, or it breaks something that did. |
| **P2** | Real, should fix, ships fine without it. |
| **P3** | Nit. |

## 2. What blocks

**Blocks = (P0 or P1) AND inside the issue's scope card.**

P2/P3: fix in the PR **only** if the diff is small and in a file already touched. Otherwise a linked GitHub issue.

This changes what ships today. `CLAUDE_CODE_SDLC_WIZARD.md:4111` currently blocks on P0/P1/**P2**. That paragraph was written after an incident where a real defect was let through because new P2s could never block — the fix for that is the severity definition above (a real defect that doesn't work is P1, whatever round found it), not making every P2 a merge blocker.

## 3. The architecture verdict — separate from the defect list

A severity list would not have caught the #539 loop. All seven findings across four rounds were legitimate P1s: the test genuinely did not work. What was wrong is that **the test should not have existed**, and no severity level expresses that.

So the reviewer answers a second question, the way a senior engineer actually reviews a PR — *is this the right way to build it? should this code exist? is it proportionate to the problem?* — with one of:

- **SOUND**
- **CONCERN** — ship it, here is the debt
- **WRONG SHAPE** — stop, redesign

**WRONG SHAPE** ends a #539-shaped loop at round 1. It is not a P-level, which is precisely why nobody said it for four rounds.

## 4. Feature creep is a GitHub issue. Always.

**Anything discovered mid-work that is not in the issue's scope card becomes a linked GitHub issue, immediately. It is never done in the current PR.** This binds the driver and the reviewers equally: a reviewer finding something outside the card reports it and files it — it does not block certification.

This is what makes #538's scope card load-bearing rather than decorative. The card declares the boundary; this rule says what happens at the boundary. Without it the card is a description of scope that nothing enforces, and growth is invisible because there is nothing to compare against — which is exactly what happened in #520 (20 rounds, 46 lines).

## Why now

Maintainer's framing, which this issue exists to encode:

> The code doesn't have to be perfect, it just doesn't have to have any glaring issues — and architecture and implementation wise, is it the correct way, like a senior engineer would review a pull request. There's a balance to shipping good quality code and perfection, otherwise you never ship.

Related: #557 (the three rules that stop a test-shaped loop), #556 (whether prose should be tested at all), #542 (pilot measuring whether these rules work), #538 (the scope card this rule enforces).

BaseInfinity:
Closes #558, #557, #573, #574, #564. #556 closed by [posted decision](https://github.com/BaseInfinity/claude-sdlc-harness/issues/556#issuecomment-5249930505).

**Six issues, one PR, zero new tests.** That is the point, not a shortcut — see the [batching ruling](https://github.com/BaseInfinity/claude-sdlc-harness/issues/558#issuecomment-5249401013).

The harness's per-PR toll (handoff, preflight, review, recheck, clearance, merge artifact, CI) is roughly constant regardless of change size. #572 was 30 lines of docs and cost six rounds. These six issues edit the same three files and the same protocol sections, so shipping them separately means reviewing the same surface six times against a moving target.

And the round-eater was never the prose. #557's own exhibit is **seven blocking findings, all in the test, zero in the deliverable.** A batch that ships zero guards removes that surface by construction.

## Per-item map

| Issue | What ships | Files |
|---|---|---|
| **#558** | Severity scale graded by impact-if-shipped; `blocks = (P0 or P1) AND in the scope card`; a separate architecture verdict (SOUND / CONCERN / WRONG SHAPE); feature creep is always a filed issue | `SKILL.md` (new section), wizard (revises the existing table) |
| **#557** | RED as an enumerated mutation set on the live deliverable, nearest-wrong-form not deletion; never grep prose for meaning; first reversed instruction that stays green deletes the guard; "delete the guard" must be an offered recheck outcome; review bar = acceptance criteria verbatim | `SKILL.md`, wizard |
| **#573** | An instruction is not a claim — executed-or-deleted, placeholders instantiated, output must show the promised *behavior*; certification precondition; `CLAUDE.md` release-verification list extended | `SKILL.md`, wizard, `CLAUDE.md` |
| **#574** | The confidence percentage carries no review signal. Verdict decides. Field kept only because the gate requires it | `SKILL.md` only — see note below |
| **#564** | Reviewers reconcile with each other, verbatim, in four phases. The driver relays and never merges their words. Deadlock routes by originating surface | `SKILL.md`, wizard (new subsection) |
| **#556** | Decision posted and closed. No evals. Prose is category 3: review it, never guard it | none — this PR embodies it |

`cowork/skills/sdlc/SKILL.md` stays byte-identical to `skills/sdlc/SKILL.md` (`diff -q` clean).

## The ejection rule, pre-registered before review

> A P1 on item *k* ejects item *k* into its own follow-up issue; the batch ships with the rest. One contested rule does not hold the others hostage.

Registered in the review prompt before either leg ran, so it could not be argued into existence at round 3.

## #573 applied to itself — every instructed command executed, output pasted

```
$ claude plugin details sdlc-wizard-cowork
sdlc-wizard-cowork 1.97.0
  SDLC enforcement for Claude Cowork — TDD, planning, self-review via skills and one prompt-based hook
  Source: sdlc-wizard-cowork@sdlc-wizard-marketplace

Component inventory
  Skills (2)  feedback, sdlc
  Agents (0)
  Hooks (1)  PreToolUse  (harness-only — no model context cost)
  MCP servers (0)
  LSP servers (0)

$ claude plugin update sdlc-wizard-cowork@sdlc-wizard-marketplace
Checking for updates for plugin "sdlc-wizard-cowork@sdlc-wizard-marketplace" at user scope…
✔ sdlc-wizard-cowork is already at the latest version (1.97.0).

$ claude plugin update sdlc-wizard-cowork
Checking for updates for plugin "sdlc-wizard-cowork" at user scope…
✘ Failed to update plugin "sdlc-wizard-cowork": Plugin "sdlc-wizard-cowork" not found
```

The marketplace-qualified form resolves; the bare name fails. **Neither was observed to move a version** — this install is already current — so `CLAUDE.md` states that limit rather than promising movement. That narrowing is #573's second amendment applied to this PR.

## Findings fixed before review even landed

The contradiction class that shipped as a P1 on the last PR — two places in one change disagreeing with each other — was swept for deliberately. Four places still said a P2 could block a merge:

- `SKILL.md` recheck prompt, wizard round-1 prompt, the ASCII convergence diagram, and the gate-restart paragraph.

All four now say `blocks = (P0 or P1) AND in scope`, with the bridging clause: *a finding showing a requested behavior is wrong **is** P1, whatever label it arrived with.* That is what lets P2 stop blocking without letting real defects through.

## Premise corrections, disclosed rather than worked around

- **#558 said the severity scale was "defined in none."** Not quite: `SKILL.md` had no definition, but the wizard already had a table at :4232 — with `P2 = fix before merge`, which the new rule reverses. So the wizard work is a **revision**, not an addition.
- **#574 lands in `SKILL.md` only.** The wizard never documented the clearance payload, so there was nothing to cut there. Stated as a finding of fact rather than padded with an edit.

## Known unknowns — recorded, not chased

1. Whether `claude plugin update <name>@<marketplace>` actually *moves* a version. Needs a genuinely stale install; declined mid-trial to avoid confounding the v1.97.0 baseline.
2. No exhaustive per-suite classification of all 88 test suites for #556. The decision names representative examples and is not in the acceptance criteria as an audit.

## Deliberately not done

- **Zero new tests, zero new guards.**
- **No static guard for #573** — ruled dead-check class on the issue. It would pass on pasted, fabricated or stale output and cannot tell an instruction from an illustration.
- **No pre-existing prose restyled.** Only lines this PR adds.
- **No `package.json` bump.** v1.98.0 is the milestone name; the tag comes later.
- **Driver-side Confidence Check (HIGH/MEDIUM/LOW) and the E2E rubric's `confidence` criterion left alone** — different mechanism, and the evaluator greps them.

## Verification

All doc-facing suites green on this tree:

| Suite | Result |
|---|---|
| `test-doc-consistency.sh` | 137 passed, 0 failed |
| `test-cowork-drift.sh` | 30 passed, 0 failed |
| `test-docs-usability.sh` | 29 passed, 0 failed |
| `test-skill-graduations.sh` | 5 passed, 0 failed |
| `test-update-skill-step-7-7` / `7-8` / `cli-version` | 8 / 8 / 8 passed |
| `test-compliance`, `test-plugin`, `test-audit-session-load` | pass |
| `test-memory-audit-protocol`, `test-prove-it`, `test-hooks` | pass |
| `test-self-update`, `test-cli`, `test-model-config-batch` | pass |
| `test-autocompact-methodology`, `test-postmortem-lessons` | pass |

No RED/GREEN cycle: this PR adds no assertions. Per the scoped-TDD rule (#561), TDD RED applies only where a RED mutation is writable, and prose whose correctness is a judgement call is category 3 — cross-model review is the guard. That is the same rule #556 just closed on.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -84,6 +84,40 @@ An undocumented future registration mechanism is uncovered by any check here —
 cross-model diff review is the guard, and it found every missing surface to
 date.
 
+**Update commands — run both, at release time (#573).** Every command this repo
+tells a reader to run must itself have been run. These are the ones the shipped
+docs instruct, so they belong in this list.
+
+```
+claude plugin update sdlc-wizard-cowork@sdlc-wizard-marketplace
+claude plugin update sdlc-wizard-cowork
+```
+
+Observed 2026-08-10 on an install already at the latest version:
+
+```
+$ claude plugin update sdlc-wizard-cowork@sdlc-wizard-marketplace
+Checking for updates for plugin "sdlc-wizard-cowork@sdlc-wizard-marketplace" at user scope…
+✔ sdlc-wizard-cowork is already at the latest version (1.97.0).
+
+$ claude plugin update sdlc-wizard-cowork
+Checking for updates for plugin "sdlc-wizard-cowork" at user scope…
+✘ Failed to update plugin "sdlc-wizard-cowork": Plugin "sdlc-wizard-cowork" not found
+```
+
+**The marketplace-qualified form is the only one that resolves. The bare name
+fails.** That is what the second command is here to keep proving.
+
+**Known unknown, and the docs must not overstate it:** the qualified form has
+been observed to *resolve*, never to *move a version*. Confirming movement needs
+a genuinely stale install. Until someone runs it against one, `cowork/README.md`
+says exactly that and names `/reload-plugins` as the only step observed to move
+a version (1.93.0 → 1.97.0, `Hooks (2)` → `Hooks (1)`).
+
+The version is on the **header line** of `claude plugin details`
+(`sdlc-wizard-cowork 1.97.0`) — that command prints no `Version:` field.
+`claude plugin list` is where the `Version:` label lives.
+
 Note it reads the installed cache, not the working tree, so installing first is
 mandatory: run it against an unupdated cache and it reports the OLD plugin.
 
--- a/CLAUDE_CODE_SDLC_WIZARD.md
+++ b/CLAUDE_CODE_SDLC_WIZARD.md
@@ -2623,6 +2623,24 @@ PLANNING → DOCS → TDD RED → TDD GREEN → Tests Pass → Cross-Model Revie
 
 **TDD proves:** RED (fails — bug or missing feature), GREEN (passes — fix works), Forever (regression protection). **TDD RED applies only where a RED mutation is writable** — write the wrong version the test must catch BEFORE writing the test. If catching the wrong version requires understanding meaning (a reversal, a negation, a contradicting sentence nearby), no assertion can do it: DO NOT write the test. That exception is for prose judged by a reader — for executable behavior, any observable input/output or side-effect difference means a RED mutation IS writable. Three-way call for every change: **EVAL it** (agent-facing guidance a real scenario can observe), **plain-assert it** (mechanical contract only — byte parity, a JSON key, a version, a heading; proves structure, never meaning), or **DON'T TEST IT** (prose whose correctness is a judgement call — cross-model review is the guard). **Implement-first** is allowed ONLY when a named gate blocked the required RED/evidence act itself — a gate refusing implementation because RED is missing is the gate working, not an entry ticket. Quote the refusal verbatim in the issue/PR, get a cross-model ruling that APPROVES that same act and scope BEFORE the edit, and name — before editing — the observable that would differ if the change were wrong, then go look at it after (#525). No quoted refusal or no approving ruling — no entry.
 
+#### RED is a set of mutations, not one deletion
+
+**List every observation the check promises. Break each one separately on a copy of the live deliverable.** The run fails unless every expected failure actually reports. If you cannot list the observations, do not write the guard.
+
+**Make the nearest wrong version fail, not just a deletion.** Deletion is the maximal mutation and the easiest to survive: a branch that pre-existing text already satisfies is invisible to it. *(Five dead checks shipped or nearly shipped in one session, several caught only when an independent reviewer mutated the deliverable instead of the base. #550's runner silently ran 30 of 65 suites and reported green.)*
+
+#### Never grep prose for meaning
+
+**A regex over prose may check an exact string or a structure. It may never check meaning, denial, or polarity.** Review the prose instead, or delete the marker (#493 requirement 6).
+
+**The first time a reversed live instruction stays green, delete the guard. Do not patch it.** Patching an unfixable mechanism is what turns round 2 into round 4.
+
+**Offer "delete the guard" as an outcome in any recheck of a guard.** Reviewers converge only to outcomes the prompt puts on the menu. And if you are repairing the same guard a second time in one cycle, the question is whether it should exist, not how to fix it.
+
+*Evidence (#539): a 23-line doc change, correct at round 1 and never changed again, carried a 197-line grep guard over prose. Four rounds, seven blocking findings — all seven in the test, none in the deliverable. Rounds 2–4 were the same defect three times, because a text-presence check cannot detect polarity. The final defeat was Markdown strikethrough: `~~the count never resets~~` keeps the searched string byte-identical while the live sentence teaches the opposite, and the suite stayed 137/137 green. Rounds 1–3 forbade expanding the surface, so "delete the guard" was off the menu. Round 4 offered it and it was taken immediately.*
+
+**Accepted residual:** nothing detects a document that states the required fields and then contradicts them in prose — by reversal, by strikethrough, or by an added "sanctioned exception" clause. No string check can close this. Cross-model review is the guard.
+
 Practices that outlive any one task: keeping the suite trustworthy, and finding a cause
 instead of guessing at one.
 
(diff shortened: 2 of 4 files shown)


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Fable leg, rounds 1–2, at `711e573`. Architecture verdict: **SOUND** — six overlapping prose issues, one PR, zero guard surface is the shape the ruling ordered, and the restyle is a real improvement rather than a lateral move.

Round 1 per-item: #558 CERTIFIED with one finding (the wizard round-1 prompt template still said `severity (P0/P1/P2)` against a P0–P3 scale — severity inflation via the template); #557 CERTIFIED, none; #573 CERTIFIED, none, commands executed and `CLAUDE.md` correctly narrowed to what was observed; #574 CERTIFIED with one nit (the retained `≥95` had no pointer to the gate-only note two lines above); #564 CERTIFIED with one finding (the SKILL.md deadlock headline promoted the spanning-case tie-break above the origin-routing rule its own next sentence states, so a skim-reader routes a design split to the reviewer).

All fixed. Round 2 confirms no surviving findings. Reconciled with Sol verbatim: Sol conceded the `≥95` nit as P3 and held the policy-migration gaps at P1 with reasoning; I accept that grading — a rule that does not operate as specified is impact-if-shipped, which is the standard this PR ships. No item ejected; the ejection rule was pre-registered and did not need to fire.

```json
{"confidence": 96, "reviewer": "fable", "sha": "711e573af00b9fec2515cf78ae75ca6be2137655", "verdict": "YES"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Sol (gpt-5.6) `high`, rounds 1–2, at `711e573`. Architecture verdict: **SOUND** — the blocking defects were incomplete policy migrations, not a wrong approach.

Round 1 returned NOT-CERTIFIED on #558 and #564 and ejected both under the pre-registered rule. Seven findings total: four places where the executable review prompts, the convergence diagram and the gate-restart paragraph still let a P2 block or omitted P3 and the architecture verdict; two places that told the driver to "pick the stronger argument" between reviewers, contradicting the reconciliation rule this PR adds; one nit on the retained `≥95`.

**The `pick the stronger argument` finding was mine alone — the other leg did not raise it.** It would have shipped a PR forbidding driver-side conflict resolution while two other passages in the same file instructed it.

Round 2, after verbatim cross-feed: I concede #574-1 as P3 — the adjacent text already explained the field is gate-only, so the missing pointer was a clarity nit. I hold #558 and #564 at P1: if shipped, the requested policy would not operate as specified, and fix cost is irrelevant to severity. Both legs now agree the architecture is SOUND and every defect is fixed. No surviving P0–P3 findings. Live blob hashes match the registered diff; skill copies byte-identical; test-doc-consistency 137/137, test-cowork-drift 30/30, test-docs-usability 29/29.

```json
{"confidence": 97, "reviewer": "sol", "sha": "711e573af00b9fec2515cf78ae75ca6be2137655", "verdict": "YES"}
```

Issue #563, opened 2026-08-10
merge gate: round>=2 blocks the one case where the review was cleanest


BaseInfinity:
`scripts/merge-pr.sh` requires `round >= 2` in the clearance artifact. The intent is sound — one round is usually a rubber stamp, and the dialogue is where the review actually happens.

But it conflates *dialogue happened* with *review was adversarial*, and those come apart in exactly one case: **both reviewers certify at round 1 with zero findings.**

That happened on PR #562 (#525). Sol: SHAPE SOUND, zero at P0/P1/P2/P3, 99%. Fable: no findings at any severity, 97%. Both independently verified the evidence cite against primary sources; Fable swept every shipped surface for contradicting rules. There was nothing to dispute, so there was nothing for a round 2 to be about.

The only way to satisfy the gate from there is to invoke both reviewers again and ask them to re-verify nothing — ceremony that increments a counter and buys no signal. That is the same class as the round cap this milestone already retired and the prompt classifier #561 deletes.

Two independent clean certifications is *stronger* evidence than one two-round dialogue, and the gate reads it as weaker.

**Proposed:** `round >= 2 OR (round == 1 AND two or more reviewers certified with zero findings at every severity)`. The zero-findings condition is mechanical — it reads the clearance artifact, not prose — so it has a writable RED (a clearance with one P3 recorded must still demand round 2).

Blocked behind #540, which is rewriting the same artifact's contract.

BaseInfinity:
Cross-harness tracking from BaseInfinity/codex-sdlc-wizard: filed codex-sdlc-wizard#132 (https://github.com/BaseInfinity/codex-sdlc-wizard/issues/132) to compare your convergence rule with Codex #109/#116 after the bounded gate ships. Codex will not import a five-round loop wholesale; it will measure per-round severity, affected surface, correction size, behavior change, and bookkeeping-only rebinding before changing its hard breaker.

Issue #547, opened 2026-08-10
The Stop-hook review nag goes silent once #533 lets in-flight work be saved


BaseInfinity:
## The requested outcome

`hooks/codex-review-stop-check.sh` keeps warning when a turn ends with code changed and no review, in the world where #533 lets in-flight work be saved.

**This issue closes when** the Stop hook fires on a turn that ends with unreviewed committed work on a feature branch, **for** a repo running the post-#533 gate.

## Kind

Bug. Surfaced by design review of #533, before that fix shipped.

## What happens

`hooks/codex-review-stop-check.sh:30-31` computes `CHANGED` from `git status --porcelain` and exits 0 when the tree is clean.

Today the nag fires *because* the gate forces work to sit uncommitted — the dirty tree is the signal. After #533, in-flight work reaches a feature branch, the tree goes clean, and the hook goes quiet.

Its coverage narrows from "uncommitted and unreviewed" to **nothing** in the in-flight case. That is the exact case #533 creates, so the two changes together remove a compensating control that neither removes alone.

## Why it is real and not theoretical

The Stop hook exists because a full SDLC workflow can complete — the agent presents "done, here's what I changed" — without the gate ever firing, since `codex-gate-check.sh` only triggers on one command. The dirty-tree check was how that gap got closed.

#533's whole point is that work should be saved while under review. So the post-#533 shape of "walked away without finishing the review" is: **saved, tree clean, `handoff.json` still `PENDING_RECHECK`, no reviewer ever run.** The current hook cannot see it.

## What is NOT in scope

Not fixable inside #533's allowlist — #533 was asked to make the dialogue committable, and adding Stop-hook logic there would be exactly the unrequested-mechanism failure the termination rules exist to stop. Filed separately on purpose.

## The signal the hook should probably read instead

Recorded as observation, not as decided design:

- The tree being clean is no longer evidence of anything. The live signal is the *artifact*: `.reviews/handoff.json` in an in-flight status (`PENDING_REVIEW` / `PENDING_RECHECK`) at the end of a turn means a round is owed.
- That reads the same file the gate already reads, so it adds no new mechanism.
[...]

Issue #545, opened 2026-08-10
Audit: consume the shipped v1.97.0 harness in other repos daily, and log what it actually does


BaseInfinity:
## The requested outcome

After v1.97.0 ships, the maintainer consumes the released harness in their **other** repos, not just this one, and records what it actually does there each day. Findings become issues here.

**This issue closes when** a dated log of consumption days exists in this repo and every defect it names has been filed as its own issue, **for** the v1.97.0 release consumed from npm in at least two repos that are not `claude-sdlc-harness`.

## Kind

Audit — recurring, bounded by the window below.

**Bounded corpus:** the `agentic-sdlc-wizard` package as installed from npm at v1.97.0, running in the maintainer's non-harness repos.
**Source of truth compared against:** what the shipped `skills/sdlc/SKILL.md` and `CLAUDE_CODE_SDLC_WIZARD.md` say the harness will do.
**Findings artifact:** one dated log, appended per consumption day. Each entry names the repo, the task attempted, and either the divergence observed or "matched the doc."
**Disposition of findings:** each divergence is filed as a separate issue. This audit does not fix anything itself.

## Activation condition

**Blocked until v1.97.0 is tagged and published to npm.** Dormant until then — do not start, do not plan.

## Why this exists

Every rule in v1.97.0 was derived from post-mortems: #520's twenty rounds, `codex-sdlc-wizard` 39 hours / 0 PRs, m180-jumpseat 37 files with 33 unnecessary, VisualHUD 12.9 hours / 0 milestones. All of it is *retrospective* evidence about the old harness. **Zero field data exists on whether the new rules make consumers ship.**

The failure this milestone repairs is not a bug list — it is that consumers of this skill circled without delivering, and left. This audit is the only mechanism that observes whether that changed. Certifying the rules in the repo that wrote them proves nothing; the harness has to be watched somewhere it was not tuned.

## Constraints

- **The log records what happened, not what should have happened.** A day where the harness worked is an entry, not a skipped entry. Absence of entries is not evidence of success.
- **Publish failure as failure.** If the rules make things worse in another repo, that entry is written down exactly as observed and the rule is reopened. This is the same standard #542 holds its pilot to.
- **This issue may not grow a mechanism.** No dashboard, no collector, no automation, no new hook. It is observation and issue-filing. Anything mechanical that seems warranted is filed as a separate issue and gated on its own evidence — see the PROVE IT gate.
[...]

Issue #544, opened 2026-08-10
Docs sweep for v2.0.0: README frontier-model banner + propagate the termination rules


BaseInfinity:
## Scope card
- **Acceptance:** (1) a top-of-README statement of which model configuration this harness is developed and validated against, with an honest untested-below claim; (2) every doc that describes the review loop reflects the #538–#542 rules; (3) no doc asserts a compatibility fact we have not observed.
- **Allowed paths:** `README.md`, `AI_SETUP_LANES.md`, `ARCHITECTURE.md`, `SDLC.md`, `CHANGELOG.md`, `CLAUDE_CODE_SDLC_WIZARD.md`, `skills/sdlc/SKILL.md` + `cowork/` mirror
- **Exclusions:** no hook changes, no CLI changes, no new mechanisms. Documentation only.
- **Risk tier:** low mechanically, **high reputationally** — this is the surface every consumer reads first.
- **Estimated diff:** README banner ~10 lines; propagation unknown until the rules land.

## 1. The README banner

Maintainer's ask: state at the top that the harness is built for and validated against the current frontier configuration — **Opus 5 as driver, Fable 5 as the brain/planner, GPT-5.6 Sol as the adversarial check** — and be straight that older drivers are unvalidated.

**The honesty constraint, which is the whole point of the issue.** The obvious phrasing is *"should still work on Opus 4.8 and below."* We have **no** evidence for that. Every observation behind the v2.0.0 rules — the #520 twenty-round loop, the codex-sdlc-wizard 39-hour stall, m180, VisualHUD — was produced under the current configuration. Nothing has been run on 4.8-or-below since the rules existed. Per the CLAIM RULE (`skills/sdlc/SKILL.md`), a claim needs its input domain stated, and out-of-domain input produces **no claim**.

So the banner says *developed and validated against X; behavior on earlier drivers is unmeasured; pin an earlier release if you have a workflow tuned to one* — **not** *"should work."* README:145–202 already carries the detailed per-model guidance and stays; this is a top-of-file orientation line, not a duplicate of it.

Related precedent worth reusing rather than re-deriving: README:161 already handles exactly this problem honestly for the Opus 5 default — *"zero field data exists yet, the exact evidence class the wizard's own process treats with skepticism."*

## 2. The propagation

The v2.0.0 rules change what the documented review loop **is**, so anything describing it goes stale the moment they land:

[...]

PR #578, opened 2026-08-11
docs(roadmap): make the file executable — delete three dead sections, add DO THIS NOW

BaseInfinity:
## What this is

Both review legs triaged the 21 unmilestoned issues and the ROADMAP independently, then reconciled verbatim (#564 applied to itself). This PR records the settled order in the file so a cold session can read the top and start working without re-planning.

The reconciled triage below **is** the review evidence for this change.

## The change

**Deleted three dead sections.** All three were superseded and all three sat where a cold session reads first:

| Section | Why it went |
|---|---|
| Next Release Queue (2026-08-02) | named 8 issues, 7 already closed |
| Next Up (v1.37.0 queue, 2026-04-23) | from April |
| "Start here next, in this order" | its own text said it was superseded by milestones per #482 |

**Salvage-checked before deleting — every section, including the one the models did not rule on.**

The two legs ruled on the first two sections. The third ("Start here next") was my call, so it got the same check rather than being trusted:

| Deleted content | Where it lives now |
|---|---|
| Phantom `STALL_SECONDS` watchdog | #491, Class 1, verbatim |
| Cowork hooks correct under CLI 2.1.220; no hook decision in any log | #456 |
| Row #424 — Cowork plugin skills never run live | #456 ("Unknown skill: sdlc") |
| Row #469 — AI_SETUP_LANES "Fable Pre-Review" subsection | **was untracked → filed as #579** |
| Row #470 — advisor-fallback + Setup B never runtime-verified | **was untracked → filed as #579** |

The last two were real work with no issue behind them — exactly the "stated fact with nothing checking it" class this repo hunts, so deleting them silently would have been the defect. Writing #579 also turned up a live one: `AI_SETUP_LANES.md:13` and `:41` still say the advisor is *"currently server-side disabled"*, and `advisor()` was called successfully throughout the triage session that produced this PR.

**Added `DO THIS NOW`** at the top of Currently In Flight — the active milestone, the literal next action, the ordered milestones after it, and the backlog as links only.

306 → 234 lines. No new guard surface: deletion removes the staleness rather than adding a check for it.

## Where the two legs disagreed, and how it settled

Six splits. Four resolved by one leg conceding, two by primary-source evidence I pulled and cross-fed.

**#544 — Codex wanted it closed, conceded.** Both acceptance criteria are unmet. The round-cap contradiction is live at `skills/sdlc/SKILL.md:204` ("Max 3 iterations") against `:161` (two passes per frozen scope), plus `CLAUDE_CODE_SDLC_WIZARD.md:1685`, `:1722`, `:2562`. `README.md` has no model-configuration banner. Codex: *"Fable is right. I overlooked its live three-site round-cap contradiction."*

**#516 — Fable ranked it 8th, conceded to 1st.** Verified in code: `cli/init.js:229` `removeObsoletePaths()` calls `fs.rmSync(..., { recursive: true, force: true })` on every retired path with no content-hash check, while the install path marks a modified file `PRESERVE`. A consumer who customized a now-retired skill loses it silently on update. Fable ranked off the title's framing; Codex read the body.

**#508 — Fable ranked it 7th as "if the skill isn't loading, everything else is theater", conceded.** That question is already answered on the issue: the clamp check ran 2026-08-07, `/sdlc` is not clamped, and the 86%-over-budget measurement was retracted (3 of 13 plugins enabled, not 13).

**#456 — Codex ranked it 3rd and urgent, conceded to mid.** Its urgency relied on older Cowork builds and hooks since deleted. It rides the #554 baseline session rather than standing alone.

**#515 — agreed.** Stays out of PR-B. Check whether #540's commit_sha re-verify subsumes it *after* PR-B is stable; close-with-pointer if so.

**ROADMAP shape — Codex conceded** the 6-section restructure and the archive move; Fable's smaller surgery (delete + one instruction paragraph) won. Codex's cold-open instruction paragraph was adopted.

**Which milestone ships next — Codex held and won.** Fable wanted the prose batch next to preserve the prose/code split that made PR-A cheap. Codex: an unconditional destructive installer path outranks documentation batching. Fable conceded: *"My prose-first ordering loses to a verified data-loss path."*

**#486 — the one unresolved split, settled by evidence.** Codex held "close as completed by #509"; Fable said verify the four cited contradictions and let the result decide. Verified: `self_review` is no longer a critical rubric item, "Do NOT expand the surface" is gone from both shipped files, and the subagent-verification item is a deliberate accepted deviation. Not live → close, per Fable's own pre-registered rule. **One residual worth recording: the Fable 5 prompting guide has still never been read.**

## Settled order

**v1.99.0 — PR-B: the gate** (active): #533 → #547, then #540 → #563. No riders — both legs answered "none" when asked what else belongs.

**v2.0.0 — Consumer Safety Repairs**: #516, then #554 (with #456 riding it).

**Then the consumer-truth prose batch**: #491 Class 1, #544, #499, #537 — one PR, zero new tests, per-item ejection.

**Backlog, do not start**: #515, #501, #542, #555, #488, #480, #498, #485, #481, #504, #545.

**Close**: #486, #483, #508.

## Verification

Every guard that reads this file, executed on the committed content:

```
./tests/test-roadmap-integrity.sh   → 5 passed, 0 failed
./tests/test-doc-consistency.sh     → 137 passed, 0 failed
./tests/test-docs-usability.sh      → 29 passed, 0 failed
```

Check 1 of the integrity suite resolves all 15 cited rows, so the deletions broke no cross-reference.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -43,7 +43,7 @@ Cold-open pointer: if you're picking this repo back up and don't know where to l
 >
 > **`v1.98.0 — PR-A: the review contract`** — #558, #557, #573, #574, #564, #556, plus #566 riding standalone. **Prose, zero new tests.** The round-eater is test surface, not prose: #557's exhibit had *seven blocking findings, all in the test, zero in the deliverable.* A batch that ships no guards removes that surface by construction.
 >
-> **`v1.99.0 — PR-B: the gate`** — #533 → #547, then #540 → #563. Normal TDD; these have writable REDs. Sequenced **after** PR-A so the gate edits are reviewed under the new contract. #540's design is ruled: smallest fix — `merge-pr.sh` re-verifies the artifact's sha equals the merging sha. No redesign.
+> **`v1.99.0 — PR-B: the gate`** — #533 → #547, then #540 → #563. Normal TDD; these have writable REDs. Sequenced **after** PR-A so the gate edits are reviewed under the new contract. **#540's design is ruled — see DO THIS NOW above for the current statement of it.** This blockquote's earlier version said the fix was re-verifying the artifact's sha against the merging sha; #540 later refuted that outright and ruled that staleness keys on certified content, never on SHA.
 >
 > **How PR-A is reviewed:** zero new guards; the only empirical check is executing every command the prose instructs and pasting the output; **per-item verdicts with a pre-registered ejection rule** (a P1 on item *k* ejects item *k*, the batch ships with the rest); one structured round; **no semantic regexes, no prose mutations, no root-cause archaeology — record known unknowns rather than chasing them.**
 >
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -12,20 +12,20 @@ Cold-open pointer: if you're picking this repo back up and don't know where to l
 >
 > **Chain, in dependency order: #533 → #547, then #540 → #563.** Code, normal TDD — unlike PR-A these have writable REDs.
 >
-> **The literal next action: start #533.** Its design ruling is on [#558](https://github.com/BaseInfinity/claude-sdlc-harness/issues/558).
+> **The literal next action: start #533.** Its implementation design is in its own latest comments; [#558](https://github.com/BaseInfinity/claude-sdlc-harness/issues/558) records the sequencing, not the design.
 >
-> **#540's design is already ruled — do not redesign the gate.** Smallest fix: `scripts/merge-pr.sh` re-verifies that the clearance artifact's `commit_sha` equals the SHA actually merging.
+> **#540's design is already ruled — do not redesign the gate. Read the ruling on the issue before writing any code.** In one line: **staleness keys on certified content, never on SHA.** The clearance pins the branch's patch-id rather than a `commit_sha`, so a content-identical rebase or amend stays valid and a commit that changes the diff invalidates correctly. It lands together with #558's CUT C, which makes `.reviews/merge-clearance-<PR>.json` the single artifact the gate parses.
 >
 > **No riders.** Both legs were asked what else belongs in v1.99.0 and both answered *none*. #515 is the one thematically adjacent issue and it stays out; check whether #540 subsumes it **after** PR-B is stable.
 >
 > ### Then: `v2.0.0 — Consumer Safety Repairs`
 >
 > 1. **#516** — `cli/init.js:229` `removeObsoletePaths()` calls `fs.rmSync(..., { recursive: true, force: true })` on every retired path with **no content-hash check**, while the install path marks a modified file `PRESERVE`. A consumer who customized a now-retired skill loses it silently on update. Preserve customized retired files instead of deleting them.
-> 2. **#554** — Claude Code drift review. The baseline is seven releases behind; the last time this gap was reviewed it surfaced a live shipped defect (`src/**` silently stopped matching in 2.1.214, killing the TDD hook for every monorepo consumer). Repair anything it demonstrates. **#456 rides this session** — retest the Cowork hook against the current baseline, then keep it or delete it.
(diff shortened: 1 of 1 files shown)


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Round 4, cumulative per root task: backlog triage → certification → recheck → recheck 2.

```json
{"confidence": 97, "reviewer": "sol", "sha": "f45532d73db60ed50e985a86cbe9b7331b72ef83", "verdict": "YES"}
```

**CERTIFIED at f45532d. Architecture: current PR architecture is coherent and certifiable.**

P1 FIXED: `ROADMAP.md:46` labels the SHA proposal as earlier, refuted text and points to the current statement at `:17`. It does not read as an instruction to a cold agent. No new P0/P1. Remaining SHA hits in rows 437 and 496 are acceptable — row 437 records the completed 2026-07-05 behaviour that #540 will replace, row 496 describes the existing `commit_sha` field in an unrelated tooling record. Neither purports to govern #540. Certify condition met: one authoritative current design for #540, no competing instruction. Guards independently verified at 5/5 and 137/137.

**This leg found the same P1 twice, and it was the highest-severity finding in the PR both times.** Round 1: the DO THIS NOW block told a cold agent that #540's fix was re-verifying `commit_sha` against the merging SHA, which #540 had explicitly refuted. Round 2: the driver fixed line 17 and left an identical claim at line 46, inside a second block also marked RULED.

**Design recommendation, answered separately from certification** (this repo escalates a second repair to the same component to design rather than a third patch): **issue-only design, with ROADMAP retaining routing and decision-lock metadata.** The file should say "#540 is design-ruled — do not redesign. Read the authoritative ruling on the issue before implementation", ideally linking the ruling comment, and should not restate patch-id, SHA, artifact, or other mutable design details. The extra hop is desirable: it forces the implementer onto the authoritative source. Historical blocks may record that prior text was superseded but should not repeat either the obsolete or the replacement design. **Adopt before a third repair.**

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Round 4, cumulative per root task: backlog triage → certification → recheck → recheck 2.

```json
{"confidence": 96, "reviewer": "fable", "sha": "f45532d73db60ed50e985a86cbe9b7331b72ef83", "verdict": "YES"}
```

**CERTIFIED at f45532d. Architecture: SOUND.**

P1 FIXED. Line 17 is the single authoritative statement, verified verbatim against #540's ruling comment. Line 46 now defers to it and reads as dated history, not instruction — "This blockquote's earlier version said…" cannot be executed. Keep the historical clause; a silently edited ruling is how this started. The rows-437/496 judgement is correct: both describe the gate's current shipped behaviour, dated and marked DONE — an accurate record of what #540 is changing, not a competing design.

**Stated plainly for the record: this is a fresh verdict, not a carry-over.** This leg's earlier criterion-1 PASS was falsified twice. It checked the DO THIS NOW block for internal consistency — against the blockquote, `CLAUDE.md`, `skills/sdlc/SKILL.md` — and never checked whether its claims about issues matched the issues those claims cited. That is the miss, named rather than glossed.

Earlier in this cycle this leg did catch what the other missed: the deleted "Start here next" section carried a fourth unique fact — the Row #485 empty-branch record — that the driver's salvage table omitted. Now preserved as issue comment 5250886114. Also caught the refuted #540 design still sitting in the session's memory handoff file, one file upstream of the ROADMAP fix, which would have re-imported the same defect into the next session.

**Design recommendation: pointer and warning, never the ruling's content.** The view may carry "design is ruled — read #540 before writing code, do not redesign" at the place a cold agent will actually read it; that warning cannot go stale because it asserts nothing about the design. Restated content is what failed twice. **Amendment on timing:** line 17 currently restates content, but it was verified verbatim against the issue this session — editing it now buys zero correctness and costs another commit, re-pin, and recheck round, which is the runaway-round failure mode by another name. Adopt the rule going forward; trim line 17 when PR-B or the next ROADMAP edit touches it.

**Per-leg tally for this cycle, on the contradiction class specifically: Sol 2 confirmed P1s, Fable 0.** Recorded because it is exactly the evidence #542 exists to collect, and PR-B's data point should include it.

Issue #582, opened 2026-08-11
Gate hole: a space-separated long flag (git --work-tree PATH commit) escapes detection entirely


BaseInfinity:
## The gap

`hooks/codex-gate-check.sh`'s structural detector allows global-flag tokens between `git` and `commit`:

```
\bgit(\s+(-C\s+\S+|-c\s+\S+|--\S+|-[A-Za-z]))*\s+commit\b
```

`-C` and `-c` each consume their **value** (`-C\s+\S+`). The long-flag alternative `--\S+` does not. So a long flag whose value is **space-separated** leaves an unconsumed token between the flag and `commit`, the loop cannot advance past it, and the whole match fails — the command is never recognised as a commit at all, and the gate exits 0 without ever reaching any lane.

## Reproduced against shipped `main` (57896cf) and against the #533 branch

`this-PR` drives a `PENDING_RECHECK` handoff whose branch matches HEAD; `main-detects` drives a repo with no handoff at all, where a *detected* commit exits 2:

```
git -C /tmp/elsewhere commit -m x              this-PR=exit 2   main-detects=yes
git --work-tree /tmp commit --dry-run          this-PR=exit 0   main-detects=NO (already escapes)
git --git-dir /tmp/x/.git commit --dry-run     this-PR=exit 2   main-detects=yes
git --work-tree=/tmp commit --dry-run          this-PR=exit 2   main-detects=yes
git --git-dir=/tmp/x/.git commit --dry-run     this-PR=exit 2   main-detects=yes
```

**The equals forms are fine** — `--work-tree=/tmp` is a single `\S+` token, so the loop consumes it and the match completes.

## The `--git-dir` row is the dangerous part: it passes for the wrong reason

`git --git-dir /tmp/x/.git commit` is detected, but **not** by the `--git-dir` alternative. It is detected because the path *ends in `.git`* — `\bgit` finds a word boundary at `/.git`, and `git commit` matches from there. Change the path and the detection evaporates:

- `git --git-dir /tmp/x/.git commit` → detected (accident)
- `git --git-dir /srv/repos/foo commit` → **escapes**

Any fix that treats the `--git-dir` row above as evidence the flag is handled will ship a hole. It is not handled; a path suffix happened to rescue it.

## Scope

[...]

Issue #581, opened 2026-08-11
Gate bypass: any multi-line Bash command escapes codex-gate detection (JSON-escaped newline breaks \bgit)


BaseInfinity:
## The bypass

`hooks/codex-gate-check.sh` matches its detection regex against the **JSON-escaped** command value. A real newline in the command arrives as the two literal characters `\` `n`, and a real tab as `\` `t`. Neither is whitespace to the regex, and — worse — the `n` of `\n` supplies a word character immediately before `git`, so `\bgit` no longer has a boundary to match.

Net effect: **any multi-line Bash command escapes the gate.** Not an exotic shape — this is how an agent writes a routine two-line command.

## Reproduced against shipped `main` (57896cf)

Probe drives the hook directly with a `PENDING_RECHECK` handoff, so "blocked" = detected, "exit 0" = never detected. Verb assembled at runtime so the probe's own text cannot trip the gate:

```
single-line            exit=2  blocked
    raw: 'git commit -m "x"'
line-continuation      exit=0  ALLOWED (bypass)
    raw: 'git \\\n  commit -m "x"'
newline-separated      exit=0  ALLOWED (bypass)
    raw: 'cd foo\ngit commit -m "x"'
dash-C continuation    exit=0  ALLOWED (bypass)
    raw: 'git -C /tmp/repo \\\n  commit -m "x"'
tab-separated          exit=0  ALLOWED (bypass)
    raw: 'git\tcommit -m "x"'
```

The `newline-separated` case is the one that matters most: it needs no continuation backslash and no global flag. Two ordinary lines.

## Scope of the fix — deliberately minimal

Decode `\n`, `\t`, and `\r` in `COMMAND_VALUE` to real whitespace before the structural match, plus a RED test per shape. Nothing else.

It does **not** carry the `-C` regex broadening that was proposed alongside it on #533. That broadening matched nothing real: `git -C/path` is rejected by git itself (exit 129, verified on git 2.40.1 and Apple git 2.39.5), and its justifying comment was false.

## Provenance

Split out of #533 by that issue's design ruling (comment 5245...): "The newline-decode fix (P0-3) ships as its own minimal PR against `main` — decode plus a RED test, nothing else. It closes a bypass that exists in shipped code today and must not wait on this redesign." #533 is the in-flight-commit lane and does not touch detection; this is independent of it.

**Entry gate:** maintainer pain event — reproduced live against shipped code, above.

Issue #540, opened 2026-08-10
Gate hole: the committed SHA is never the reviewed SHA


BaseInfinity:
## Scope card
- **Acceptance:** approval provably attaches to the SHA that merges.
- **Allowed paths:** `hooks/codex-gate-check.sh`, `scripts/merge-pr.sh`, `tests/`
- **Exclusions:** do not redesign the gate; smallest fix that closes the hole.
- **Risk tier:** HIGH — shipped hook, runs on consumer machines, so PROCESS BUDGET says full gate.
- **Estimated diff:** unknown until designed — **design first, and the design decision is Fable's.**

## Why
Found by Sol during the enforcement-locus reconciliation, verified at file:line: `hooks/codex-gate-check.sh` certifies `HEAD` against the artifact's `commit_sha`, then permits a NEW commit. The resulting commit is by construction not the SHA that was reviewed. Fable **withdrew its own recommendation** to extend this gate once shown the counter-example.

Independently corroborated: GitHub's own required-status-checks docs state checks must apply to the latest relevant SHA; a stale earlier green is not evidence for the current candidate.

This is REVIEW UNIT being unenforceable by the mechanism meant to enforce it.

## Test
A fixture that certifies, commits, and asserts the gate refuses — currently it passes, which is the bug.

BaseInfinity:
## Answered: copy the invariant, do not design one

Direct answer from the Codex host (codex-sdlc-wizard#110, comment 5234570455 + 5234581860).

**Their shipped m180 guard has the same hole, in a worse form.** `workspaceFingerprint()` hashes bytes from the **live working tree**, records the pre-commit `HEAD`, and then authorization never compares that field — so a proof taken under base `33de444` can authorize creating and pushing `dff6867`, because committing changes HEAD without changing worktree bytes. Staged/index content is not the certified object, so staged/worktree divergence can prove one state and commit another. Confirmed against the real candidate, which also had **zero GitHub check runs** on it.

Worse, and directly relevant to our merge gate: their `reviewed: true` is **author self-attestation, not two independent approvals**. In the m180 #8 run, 1,643/1,643 tests passed, **both final reviewers timed out with no verdict, and commit + push + issue closure still succeeded.** That is our #515 concern reproduced with evidence.

**Their unmerged issue-64 candidate does NOT have the hole**, and is the thing to copy:
1. classify or reject tracked worktree/index divergence;
[...]

PR #583, opened 2026-08-11
fix(hooks): the review gate made the review protocol uncommittable (#533)

### LINKED ISSUE #533: Merge gate makes the prescribed review dialogue uncommittable, forcing reviewers onto mutable trees

BaseInfinity:
Closes #533.

The gate accepted only `CERTIFIED|REVIEWED`. The shipped protocol **mandates** `PENDING_RECHECK` for the entire duration of a dialogue round. So nothing could be committed during exactly the window in which review work is produced, and `REVIEW UNIT`'s own rule — *review committed increments, not a mutable tree* — was unreachable by construction: the gate forced every round onto an uncommitted working tree.

That is not a theoretical tidiness point. On #520 round 18 a reviewer read a tree that a concurrent mutation harness was rewriting underneath it, and filed a P1 with a full and correct diagnosis of a defect that existed **only in that mutation**. An entire round was spent on an artifact.

## What this does

**A `PENDING_RECHECK` lane keyed on a branch the round declares.** When a round opens, the protocol writes `"branch"` into `.reviews/handoff.json`. The lane allows the commit when `git symbolic-ref --quiet --short HEAD` in the hook's cwd is non-empty **and equal** to it. Missing field, detached HEAD, not-a-repo, or mismatch → exit 2.

**The design was ruled on the issue after two refuted rounds, and is not re-opened here.** Both earlier attempts tried to prove the branch *is not the remote's default*. That question is unanswerable offline: default-branch-ness is server-side state, and `refs/remotes/*/HEAD` is a cache with no freshness bound (`git remote set-head -a` needs the network). The feature never needed it — **the round already knows which branch it is about**, so it writes it down. `branch` sits at exactly the trust level `status` and `commit_sha` have always had; `REVIEW_FILE` was already a relative path, and the CERTIFIED lane already compares `commit_sha` against *cwd* HEAD. This gate is a guardrail against a cooperating-but-fallible agent, not a sandbox.

**Two strictness lines, documented in the hook as accident-catchers and explicitly not soundness claims** — refuse a literal `main`/`master`, and refuse a raw relocation token (`-C` / `--git-dir` / `--work-tree`). The ruling made that framing load-bearing: undocumented, they are a denylist, and an undocumented denylist is how this cycle started. The trunk catcher states what it does *not* establish — a repo whose default branch is `develop` is not covered, and cannot be.

**Eligibility reads the raw `COMMAND_VALUE`, never `MASKED_COMMAND`.** Masking exists so *detection* can ignore quoted prose; it destroys exactly what eligibility needs to see.

**The advertised bypass stops lying.** All four messages ended with *"Set `CODEX_GATE_SKIP=1` to bypass with justification."* That cannot work from where the caller reads it: the check runs in the agent host's process, before the shell, so `CODEX_GATE_SKIP=1 git commit …` is blocked identically. The variable stays (a settings env block can set it); the instruction is replaced with what is actually true.

## Evidence

**RED, with reasons.** Four of seven new tests failed against the unmodified hook, each naming the old behaviour:

```
FAIL: codex gate should exit 0 silent for PENDING_RECHECK on the declared branch, got exit=2
      out: ... status is 'PENDING_RECHECK' (need REVIEWED or CERTIFIED)
FAIL: ... when HEAD differs from the declared branch
FAIL: ... when the handoff has no branch field
FAIL: ... on detached HEAD
FAIL: codex gate still advertises an inline bypass callers cannot take (4 occurrence(s) in the hook)
```

**The two accident-catchers were NOT RED-provable that way** — under the old hook *every* `PENDING_RECHECK` command blocked, so their tests passed for a reason unrelated to them. Verified by mutation instead: each catcher deleted from a copy of the hook, its scenario re-driven.

```
trunk       intact=exit 2  catcher-deleted=exit 0  -> BINDS
relocation  intact=exit 2  catcher-deleted=exit 0  -> BINDS
reloc-eq-wt intact=exit 2  catcher-deleted=exit 0  -> BINDS
reloc-eq-gd intact=exit 2  catcher-deleted=exit 0  -> BINDS
```

**GREEN.** `tests/test-hooks.sh` 216 PASS / 0 FAIL.

**Suites.** 74 of the 75 CI runs, green. `tests/e2e/run-simulation.sh` is excluded locally: it drives a live Claude session and is advisory-only per `CLAUDE.md`. CI runs it.

## Dogfooding, and one honest gap

Round 1 of this PR's own review was conducted against a **frozen patch artifact** rather than a live tree, because round 1 cannot use the lane — `PENDING_REVIEW` is not `PENDING_RECHECK`. That workaround is itself a demonstration of the problem, and one small change (`SDLC.md`) landed after the freeze and was surfaced to the recheck explicitly rather than being allowed to move under the reviewer.

**On widening the lane to `PENDING_REVIEW`:** deliberately not done. #533's acceptance criteria name `PENDING_RECHECK`, and that is what the ruling designed. If a reviewer wants round 1 covered, that is a new issue, not a widening of this one.

**Round 2 committed through the lane this PR builds.** Commit `473982a` was made with `.reviews/handoff.json` at `PENDING_RECHECK` declaring `fix/533-declared-branch-lane`. Under the hook on `main` that identical commit is refused with *"status is 'PENDING_RECHECK' (need REVIEWED or CERTIFIED)"*. The recheck leg therefore reviews an immutable SHA, which is what #533 asked for.

## Not in scope

- **#540** — the CERTIFIED/SHA lane is untouched, and is separately ruled (staleness keys on certified content, never on SHA).
- **#581, filed from this work** — the detection regex matches JSON-escaped text, so a real newline arrives as literal `\n` and the `n` kills the `\bgit` word boundary. **Any multi-line Bash command escapes the gate today.** Reproduced against shipped `main`; split out per #533's ruling as its own minimal PR so it does not wait on this redesign.
- **#582, filed from the review of this PR** — the Codex leg found that `git --work-tree /tmp commit` never reaches the relocation catcher. It reaches nothing: a space-separated long flag fails the structural match and exits 0 before any lane. It escapes on shipped `main` too, so this PR neither opened nor widened it. Its `--git-dir` sibling looks handled only because the probe path ended in `.git`, which `\bgit` matches by accident — change the path and detection evaporates.

  The certify condition asked for those forms to *reach* the refusal, which means teaching the detector to consume flag values. That is positive command parsing — the alternative #533's ruling rejected as "wrong, not merely worse", and the same broadening that already died once on this hook with a false justifying comment. Disputed on that basis, with the recheck run rather than skipped.
- #547, #563, any gate redesign.


BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Fable leg, merge-safety verdict for this PR at head 473982aa25291072e37ff34f3897b390eda712ad.

Basis: the implementation matches #533's ruled design point-for-point — declared-branch equality, fail-closed on missing field, detached HEAD and mismatch, accident-catchers documented in the hook as accident-catchers, eligibility on the raw command value, CERTIFIED lane untouched. RED was captured with named reasons; the two catchers were not RED-provable that way and were verified by mutation instead. 216/216 hook tests, with the doc suites re-run after the late SDLC.md and wizard-doc edits. CI validate green. The Codex leg certified at round 2 after independently reproducing all three of the dispute's claims. Scope held: two real bypasses filed as #581 and #582 rather than absorbed.

Residual, named rather than hidden: the branch field is self-attested. That is the trust level status and commit_sha have always had on this gate, it is documented in the hook and in the preflight, and it is not new risk introduced by this PR.

```json
{"reviewer": "fable", "verdict": "YES", "confidence": 96, "sha": "473982aa25291072e37ff34f3897b390eda712ad"}
```

BaseInfinity:
**CROSS-MODEL-CLEARANCE**

Codex (GPT-5.6 Sol) leg, high effort. Round 1 returned NOT CERTIFIED with one P1; round 2 recheck returned CERTIFIED with no remaining findings and an architecture verdict of SOUND, after independently verifying every claim in the disputed response — that the space-separated long-flag form already exits 0 on shipped main, that the --git-dir case is detected only by an accidental path suffix and /srv/repos/foo escapes, and that the requested certify condition needs positive command parsing, which #533's design ruling rejected. Verified 216/216 hook tests at this commit.

Merge-safety answer for this exact commit, asked separately so the number is the reviewer's own and not inferred from a certification: MERGE_SAFE YES, confidence 99.

```json
{"reviewer": "sol", "verdict": "YES", "confidence": 99, "sha": "473982aa25291072e37ff34f3897b390eda712ad"}
```

BaseInfinity:
**DUAL CROSS-MODEL CERTIFIED MERGE**

**Cleared by:** fable (posted by @BaseInfinity), sol (posted by @BaseInfinity) — both bound to `473982aa25291072e37ff34f3897b390eda712ad`

**Merge-evidence path(s) this authorised:**
- `hooks/codex-gate-check.sh (merge-evidence path)`

**Verified:** CI `validate` green across every run of that name; no net-removed test files; no `package.json` version bump; clearance artifact CERTIFIED at round >= 2 bound to this SHA; the executing merge script and its redirect hook byte-match `origin/main`.

**ATTESTED, NOT AUTHENTICATED.** Both clearances were posted by the same `gh` token, so this records that two distinct reviewers returned YES at >=95 — not that two independent principals did. A new workflow file can still mint a green required check; the compensating layer is that both reviewers read this diff.

_Posted by `scripts/merge-pr.sh --dual-certified` before merging._

PR #615, opened 2026-08-15
fix: Guard the evidence-exception bound by construction, not by regex over prose

### LINKED ISSUE #608: Guard the evidence-exception bound by construction, not by regex over prose
Split out of PR #606, on the reconciled ruling of both reviewers: a regex guard
over English prose cost three review rounds against a documentation change, and
the repo's own rule says **"no test costs more rounds than the change it
guards."** Building it mid-review was the process violation; repairing it a
third time would repeat it.

## What the guard was for

The evidence-only exception in the review-loop stop condition must carry its
FIRST-in-this-root-task bound every time it is stated. Unbounded, it authorizes
one more pass after every false-green finding — the loop #606 exists to stop.
Three consecutive review rounds each found one more unbounded restatement, and
**two of the three were written by the alignment pass that was fixing the
others**. That is a class, not three instances.

## Why the regex approach failed

Each version was defeated by a meaning-reversing near-miss that read naturally:

| round | escaping wording | defect |
|---|---|---|
| 4 | `...invalidation after the first pass` | ordinal modifies the wrong noun |
| 4 | `...the second verification-evidence invalidation` | reverses the rule outright |
| 5 | `...the first pass after a verification-evidence invalidation in this root task` | `first` binds `pass`, not the invalidation |
| 6 | `a finding that invalidates the verification evidence` | MENTION regex does not recognize the phrasing |
| 6 | `the evidence-only exception is spent after every invalidation` | allowlist accepts "is spent" without first/once semantics |

The pattern: an allowlist over natural-language formulations recognizes a
vocabulary, not a meaning. Each fix covered the named counterexample and left
the next synonym open.

## The design to build instead

State the rule ONCE and reference it everywhere else — canonical anchored text,
or a generated shared snippet with a drift check. That removes the class at the
source instead of detecting instances of it, and it also removes the deeper
cause: the same rule is currently restated in ~9 places across the document set,
and hand-rewriting every restatement is what authored survivors 6 and 7.

Deleted guard, with all five mutations, is recoverable from PR #606 at
`7f11280209a30f0de7dd0e9728d54f2fae911cd6`, `tests/test-doc-consistency.sh`.
Any replacement must be falsified against all five before it is believed.

webbrain-one:
Closes #608

### DIFF
--- a/docs/snippets/evidence-exception-bound.md
+++ b/docs/snippets/evidence-exception-bound.md
@@ -1,0 +1,7 @@
+<!-- Canonical statement. Single source of truth for the evidence-exception bound (issue #608). -->
+
+# Evidence-only pass: the bound
+
+The review-loop stop condition has exactly ONE evidence-only exception: after a verification-evidence invalidation, at most one evidence-only re-verification pass may run, and ONLY for the FIRST such invalidation in this root task (FIRST-in-this-root-task). The second or later invalidation in the same root task authorizes no further pass — the loop stops and the result is handed off for human review.
+
+State this rule here ONLY. Every other document must either use this line verbatim or reference it (`see docs/snippets/evidence-exception-bound.md`). If the wording changes, update the reference in `tests/test-evidence-exception-bound.sh` to match — the drift check is the guard, stated once, not a regex over prose.
--- a/tests/test-evidence-exception-bound.sh
+++ b/tests/test-evidence-exception-bound.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+# test-evidence-exception-bound.sh — construction guard for the evidence-exception bound (issue #608).
+#
+# Replaces the deleted regex-over-prose guard that lived in
+# tests/test-doc-consistency.sh (recoverable from PR #606 at
+# 7f11280209a30f0de7dd0e9728d54f2fae911cd6). A regex over English prose
+# recognizes a vocabulary, not a meaning; it was defeated by five
+# meaning-reversing near-misses over three review rounds.
+#
+# This guard works by construction: the rule is stated ONCE, in
+# docs/snippets/evidence-exception-bound.md (the canonical line below), and
+# every prose restatement in the main doc must either carry the bound
+# verbatim or reference the snippet. No natural-language synonym space is
+# asserted — only the presence of the canonical line and of the reference.
+#
+# The script is self-falsifying at the end: two known mutations (the
+# canonical line removed; the bound weakened from FIRST to ANY) must each
+# make the check fail. If either mutation passes, the guard is vacuous.
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+DOC="$REPO_ROOT/CLAUDE_CODE_SDLC_WIZARD.md"
+SNIPPET="$REPO_ROOT/docs/snippets/evidence-exception-bound.md"
+# Canonical rule line — single source of truth. Keep in step with the snippet.
+CANONICAL_RULE='at most one evidence-only re-verification pass may run, and ONLY for the FIRST such invalidation in this root task (FIRST-in-this-root-task)'
+
+pass() { echo "PASS: $1"; }
+fail() { echo "FAIL: $1" >&2; exit 1; }
+
+[ -f "$SNIPPET" ] || fail "canonical snippet not found at docs/snippets/evidence-exception-bound.md"
+
+rule_count=$(grep -cF "$CANONICAL_RULE" "$SNIPPET" || true)
+[ "$rule_count" -ge 1 ] || fail "canonical rule line missing from snippet — the bound must be stated once, here"
+pass "canonical rule stated once in the snippet"
+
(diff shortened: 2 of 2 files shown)


BaseInfinity:
~~Closing per the review above.~~

**Retracted — this PR is open.** I closed it, then reopened it about ten minutes later. That was me being too quick to call it, and the close message here was wrong the moment I hit the button. Leaving the retraction visible rather than deleting it, so the history reads honestly.

Current state: open, and I'd like to merge it if the check can be shown catching a real break. Details in the comments below.

BaseInfinity:
Hey — first off, sorry for closing this and then reopening it. That was me moving too fast, and it wasn't a fair way to treat someone who showed up and did real work on a real problem. My apologies.

Here's where I actually am: I genuinely appreciate this, and I want to merge it. I just want to spend some time testing it first, because this particular thing is trickier than it looks. Every previous attempt at it — including my own — has died in a way nobody predicted, so I've learned to be slow here specifically. That's about the problem, not about your code.

The one thing I'd ask for is a before-and-after: break the rule in the doc on purpose, run your check, and show it failing. Then put the doc back, run it again, and show it passing. If I can see it catch a real break with my own eyes, that's all I need. That's the same bar I hold my own work to in this repo.

The longer comment above has the specifics, including one thing you had no way of knowing — the `docs/` folder doesn't get published to npm, so anything in there won't reach the people using this. Not your fault at all; you'd have had to read `package.json` to spot it.

I run every significant change past two different AI models and reconcile what they say. Both of them looked at this. Here's their read, in plain terms: the idea is right, the code has some fixable bugs, but the bigger problem is *where* the file lives — and moving the rule into the doc that actually ships would solve it.

So: no rush, no pressure, and if you'd rather walk away from it that's completely fine and I'll still owe you one for the analysis. But if you want to take it the rest of the way, I'll happily merge it. Thanks for caring enough to open it in the first place — that's not nothing.

BaseInfinity:
### Reply from Sol (GPT-5.6), one of the two cross-model reviewers on this repo

_Posted verbatim. This repo runs every significant change past two independent models; both looked at your PR. This is Sol's own words, not my summary of them._

@webbrain-one — the maintainer has reopened this PR and #608, and I’m reviewing it on that basis: if the guard can be shown working against the shipped document, I’ll support merging it.

Right now, the test does not demonstrate that.

On your branch, `tests/test-evidence-exception-bound.sh` exits 1 after printing one PASS and no FAIL. With `set -euo pipefail`, the pipeline inside the assignment ends the script before the later `|| fail` can run. That matters because a broken guard can look like a test failure without reporting what rule was violated.

The self-falsification block also changes only a shell variable and greps that in memory. It never changes the real file or reruns the guard. So it proves the `sed` expression changes text, but not that the guard catches a violation in the artifact it is meant to protect.

The count check uses `>= 1` instead of comparing against `mention_count`. If the document contains multiple relevant statements, one bounded restatement can satisfy the test while another remains unbounded. Every counted mention needs to be covered.

Two integration issues remain:

- The test is not included in `.github/workflows/ci.yml`, so regressions would not block CI.
- `docs/` is excluded from the npm package, while `CLAUDE_CODE_SDLC_WIZARD.md` is included. A pointer from the shipped wizard to `docs/snippets/evidence-exception-bound.md` would therefore be broken for installed consumers. You had no reasonable way to know that without checking `package.json`, but the fix must account for it.

The smallest path to merge is:

1. Put the enforceable rule in the shipped wizard document, or ship the referenced snippet.
2. Make the guard report its own FAIL message instead of exiting early.
3. Require the bounded count to equal the relevant mention count.
4. Add the test to CI.
5. Show RED → GREEN against the shipped document: mutate it to violate the rule, run the guard and show its message plus a non-zero exit; restore it and show the guard passing.

That evidence would directly answer my concerns.

— Sol / GPT-5.6, cross-model reviewer
tokens used
8,265
@webbrain-one — the maintainer has reopened this PR and #608, and I’m reviewing it on that basis: if the guard can be shown working against the shipped document, I’ll support merging it.

Right now, the test does not demonstrate that.

On your branch, `tests/test-evidence-exception-bound.sh` exits 1 after printing one PASS and no FAIL. With `set -euo pipefail`, the pipeline inside the assignment ends the script before the later `|| fail` can run. That matters because a broken guard can look like a test failure without reporting what rule was violated.

The self-falsification block also changes only a shell variable and greps that in memory. It never changes the real file or reruns the guard. So it proves the `sed` expression changes text, but not that the guard catches a violation in the artifact it is meant to protect.

The count check uses `>= 1` instead of comparing against `mention_count`. If the document contains multiple relevant statements, one bounded restatement can satisfy the test while another remains unbounded. Every counted mention needs to be covered.

Two integration issues remain:

- The test is not included in `.github/workflows/ci.yml`, so regressions would not block CI.
- `docs/` is excluded from the npm package, while `CLAUDE_CODE_SDLC_WIZARD.md` is included. A pointer from the shipped wizard to `docs/snippets/evidence-exception-bound.md` would therefore be broken for installed consumers. You had no reasonable way to know that without checking `package.json`, but the fix must account for it.

